PR #18, opened 2025-10-30
feat: create initial planner functionality

florentianayuwono:
### Overview

Create initial planner functionality. Includes: POST create flavor, observability, and tests.

### Rationale

Needed as per spec.

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`, `documentation`).

### DIFF
--- a/internal/planner/server_test.go
+++ b/internal/planner/server_test.go
@@ -2,7 +2,7 @@
  * Copyright 2025 Canonical Ltd.
  * See LICENSE file for licensing details.
  *
- * Unit tests for the planner server and telemetry.
+ * Unit tests for the planner server.
  */
 
 package planner


planner/cmd/planner/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/canonical/github-runner-operators/internal/auth"
+	"github.com/canonical/github-runner-operators/internal/database"
+	"github.com/canonical/github-runner-operators/planner"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+const dbURI = "PLANNER_DATABASE_URI"
```
yanksyoon: I think we intend to use the 12-factor application which has the following environment variables for database:
https://documentation.ubuntu.com/charmcraft/latest/howto/manage-web-app-charms/integrate-web-app-charm/#integrate-with-a-database

They are also common database environment variables, so we could use it instead of `PLANNER_DATABASE_URI` :)

planner/cmd/planner/main.go
```diff
@@ -0,0 +1,145 @@
...
+const adminToken = "PLANNER_ADMIN_TOKEN"
```
yanksyoon: Same for these environment variables, we are probably going to use the Go 12-factor application. For such custom environment variables, refer to this document here: https://documentation.ubuntu.com/charmcraft/latest/reference/extensions/go-framework-extension/#charmcraft-yaml-config-options

> In case you want to add extra configuration options, any option you define will be used to generate environment variables; a user-defined option config-option-name will generate an environment variable named APP_CONFIG_OPTION_NAME where the option name is converted to upper case and dashes are converted to underscores.

yanksyoon (commented):
Some comments, I think we might want to architect this a little bit with dependency injections :)

planner/server.go
```diff
@@ -0,0 +1,145 @@
...
+	mux     *http.ServeMux
+	store   FlavorStore
+}
+
+// NewServer creates a new Server with the given dependencies.
+func NewServer(store FlavorStore, metrics *Metrics) *Server {
+	s := &Server{store: store, mux: http.NewServeMux(), metrics: metrics}
+
+	// Register routes
+	s.mux.Handle(createFlavorPattern, s.AuthMiddleware(http.HandlerFunc(s.createFlavor)))
```
florentianayuwono: for now can leave it without authentication

internal/planner/server.go
```diff
@@ -0,0 +1,146 @@
...
+		Name:            flavorName,
+		Platform:        req.Platform,
+		Labels:          req.Labels,
+		Priority:        req.Priority,
+		IsDisabled:      req.IsDisabled,
+		MinimumPressure: req.MinimumPressure,
+	}
+
+	err := s.store.AddFlavor(r.Context(), flavor)
+	if err == nil || errors.Is(err, database.ErrExist) {
```
weiiwang01: For `database.ErrExist`, I think it's better to return HTTP 409 Conflict, because the flavor stored in the database may differ from the provided one.

internal/planner/telemetry.go
```diff
@@ -0,0 +1,122 @@
...
+
+	return m
+}
+
+// ObserveFlavorPressure updates the gauge value for a given platform and flavor.
+func (m *Metrics) ObserveFlavorPressure(platform, flavor string, value int64) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.pressures[[2]string{platform, flavor}] = value
```
weiiwang01: The problem with caching the pressure value in memory is that there may be multiple instances of the planner server. When a flavor change hits one server, it updates the pressure metric for that server only and does not affect the others.

Instead, I think it would be better to read the pressure value from the database inside the metrics callback, since performance shouldn't be an issue, metrics scraping typically occurs only every 5–60 seconds.

internal/planner/server.go
```diff
@@ -0,0 +1,146 @@
...
+		s.updateMetrics(r.Context(), flavor)
+		w.WriteHeader(http.StatusCreated)
+		return
+	}
+
+	http.Error(w, fmt.Sprintf("failed to create flavor: %v", err), http.StatusInternalServerError)
+}
+
+func (s *Server) updateMetrics(ctx context.Context, flavor *database.Flavor) {
+	pressure, err := s.store.GetPressures(ctx, flavor.Platform, flavor.Name)
```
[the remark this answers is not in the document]
weiiwang01: Yes, I don't think we need to optimize this right now. Just implement it as in the metric scrape callback, retrieve all pressures from all platforms in the database and set the gauge value. I don't think the performance issue is severe enough to require caching at this time.

.github/workflows/planner_tests.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: 1.25
+
+      - name: Build and Test
+        env:
+          POSTGRESQL_DB_CONNECT_STRING: postgres://postgres:postgres@localhost:5432/postgres
+          POSTGRESQL_DB_PORT: 8080
```
yhaliaw: Is this DB port correct? Since other part is referring to the 5432 port.

florentianayuwono: i think this has the same purpose as APP_PORT defined in webhook tests. I was trying to follow https://documentation.ubuntu.com/charmcraft/latest/howto/manage-web-app-charms/integrate-web-app-charm/#integrate-with-a-database hence the naming

internal/planner/server.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// ServeHTTP routes incoming HTTP requests to the appropriate handler methods.
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	s.mux.ServeHTTP(w, r)
+}
+
+// AuthMiddleware is an HTTP middleware that checks for a valid user token
+// in the Authorization header.
+func (s *Server) AuthMiddleware(next http.Handler) http.Handler {
```
cbartz: I think we agreed to not add this initially (we have a dedicated user story for authentication: https://warthogs.atlassian.net/browse/ISD-4479). May I ask how we add the auth token to the database? Would this need a manual database operation?

internal/planner/integration_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	if !found {
+		t.Fatalf("flavor %q not found in db", flavorName)
+	}
+
+	// Cleanup
+	_ = db.DeleteFlavor(ctx, platform, flavorName)
+	_ = db.DeleteAuthToken(ctx, tokenName)
+}
```
cbartz: Can we reduce the amount of lines of code for the test by putting subfunctionality in functions? E.g. , every block you added a comment could live in a separate function. E.g. 

```
func applyMigrationsAndContext(ctx, uri) {
	if err := database.Migrate(ctx, uri); err != nil {
		t.Fatalf("migrate failed: %v", err)
	}
	db, err := database.New(ctx, uri)
	if err != nil {
		t.Fatalf("db connect failed: %v", err)
	}
}
```

planner/main_test.go
```diff
@@ -0,0 +1,32 @@
+package main
+
+import (
+	"net/http"
+	"testing"
+	"time"
+)
+
+// waitForHTTP keeps trying a POST request until the server responds
```
cbartz: can we use the arrange act assert pattern for the comments?

planner/main_test.go
cbartz: As mentioned above, I would move integration tests here.

cbartz (commented):
Thank you! Some comments.

internal/planner/server.go
florentianayuwono: ok i will delete this scope! for adding auth token i use this operation inside the integration test
```
	// Create an auth token to authenticate requests
	tokenName := "it-planner-" + time.Now().Format("20060102-150405.000000000")
	token, err := db.CreateAuthToken(ctx, tokenName)
	if err != nil {
		t.Fatalf("create auth token failed: %v", err)
	}
	tokenHex := hex.EncodeToString(token[:])
```

planner/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+		"platform": "github",
+		"labels":   []string{"self-hosted", "amd64"},
+		"priority": 42,
+	}
+	b, err := json.Marshal(body)
+	if err != nil {
+		t.Fatalf("marshal payload: %v", err)
+	}
+
+	resp, err := http.Post("http://localhost:8080/api/v1/flavors/it-flavor", "application/json", bytes.NewReader(b))
```
cbartz: instead of it-flavor use a random string

cbartz (commented):
Great ! Some comments!

internal/planner/telemetry.go
```diff
@@ -0,0 +1,104 @@
...
+								attribute.String("flavor", f.Name),
+							),
+						)
+					}
+				}
+				return nil
+			},
+			flavorPressureMetric,
+		),
+	)
```
yanksyoon: @copilot This function is quite nested, can this be refactored?

yanksyoon (approved):
Looks good! Just minor nits that copilot can deal with :)

internal/planner/telemetry.go
```diff
@@ -0,0 +1,104 @@
...
+	meter                = otel.Meter(pkg)
+	logger               = telemetry.NewLogger(pkg)
+	flavorPressureMetric = must(
+		meter.Int64ObservableGauge(
+			flavorPressureMetricName,
+			metric.WithDescription("The current pressure value for each flavor"),
+			metric.WithUnit("{pressure}"),
+		),
+	)
+)
```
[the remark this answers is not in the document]
yanksyoon: I think constants like `pkg`, `flavorPressureMetric` is fine to keep as a package scoped variable. However, if something requires initialization like `otel.Meter` and `telemetry.NewLogger`, depending on the contents of the initialization, it might be clearer to create an explicit dependency on the `Metrics` class to track it's lifecycle.
But I haven't looked into the contents of `otel.Mter` and `telemetry.NewLogger` yet.

internal/planner/server_test.go
```diff
@@ -0,0 +1,210 @@
...
+					}
+					if string(kv.Key) == "flavor" && kv.Value.AsString() == flavor {
+						hasFlavor = true
+					}
+				}
+
+				if hasPlatform && hasFlavor {
+					found = true
+					break
+				}
```
yanksyoon: @copilot If the target value was found, shouldn't it break from the whole nested loops? This only breaks on the innermost loop. Please fix.

internal/planner/telemetry.go
```diff
@@ -0,0 +1,127 @@
...
+		observer.ObserveInt64(
+			flavorPressureMetric,
+			int64(v),
+			metric.WithAttributes(
+				attribute.String("platform", platform),
+				attribute.String("flavor", f.Name),
+			),
+		)
+	}
+}
```
cbartz: It seems explicit unit tests are missing for these methods (though these methods are tested via the unit tests in server_test.go).

internal/planner/telemetry_test.go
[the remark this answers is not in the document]
yanksyoon: I do think there is a value in testing the input/output state with what we're doing for the interfaces that we expose (NewMetrics), as @florentianayuwono mentioned. Perhaps we could have that in the integration test? This would allow us to test the integration of two components, the library for collecting metrics (provided by otel.Meter) and our work (collectFlavorPressure).

On a different note, when we think about the purpose of a unit test, scoping the function to it's "unit" of work might be more suited for its' purpose. Here, the unit of work that provides the value of this package seems like `collectFlavorPressure`, `collectPlatformPressure`, ...etc and such.
image

As of expected future changes, what do you think about adapting the test as we shift what the value of the function provides?

yanksyoon (commented):
We're almost ready to merge! Thanks for the changes!

internal/planner/telemetry_test.go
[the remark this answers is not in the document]
florentianayuwono: it's just that the current private functions are very dependent on each other: `NewMetrics` initialize the metrics and register callback by calling `collectFlavorPressure`, and `collectFlavorPressure` call `collectPlatformPressure`, and `collectPlatformPressure` call `extractFlavorNames` and `observeFlavorPressures`. i was worried that if we want to unit test each of these functions, we will need to introduce lots of mockings for the called private functions, which will result in larger codebase. and moreover since these are private functions, im worried that introducing a lot of tests and mockings for these will result in brittle tests that need to be changed a lot everytime we make changes in the future.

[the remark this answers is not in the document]
cbartz: I had exactly in mind what Yanks visualised. I overlooked that most of these functions are meant to be private though, this is a good point.  `NewMetrics` seems to be the only exported public function. Usually, imho, we should achieve black box tests by testing the public API (unless there is complex logic in private implementation that needs specific explicit testing, e.g. complicated error paths).

Indeed, it might make more sense to use kind of "scenario" test as you have done for this particular source file. 

I think we need more alignment on our testing approach in this project in general, but I am good to go with the current approach you took for now.

cbartz (approved):
Thanks! Great work!

PR #35, opened 2025-11-18
chore: Print out coverage

cbartz:
### Overview

1. Move database integration test to tests.yaml , as it is an independent internal package
2. Print coverage summary.
3. Update status checks.

Next step would be to enforce min coverage, but the team needs to align on min coverage to be achieved beforehands.

### Rationale

1. At the moment, the database internal package execution is tight to the workflow of the webhook gateway (charm) integration tests. But as internal package, it should be tested independently. Also, we need the execution of this package for our coverage report.
2. Print out a more detailled coverage report indicting missing coverage per function.
3. Test workflow name has been updated to be more explicit. Also, distinguish between webhook-gateway and planner integration status checks.

### DIFF
--- a/.github/workflows/internal_tests.yaml
+++ b/.github/workflows/internal_tests.yaml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   test:
     name: Test Internal Packages
-    runs-on: [ubuntu-latest]
+    runs-on: [self-hosted-linux-amd64-noble-edge]
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/planner_tests.yaml
+++ b/.github/workflows/planner_tests.yaml
@@ -28,7 +28,7 @@ jobs:
   integration-test:
     needs: detect-changes
     if: needs.detect-changes.outputs.planner == 'true'
-    runs-on: [ubuntu-latest]
+    runs-on: [self-hosted-linux-amd64-noble-edge]
     name: Run Integration Tests
     steps:
       - uses: actions/checkout@v6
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yaml
```diff
@@ -9,8 +9,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  unit-test:
-    name: Run Unit and Lint Tests
+  test:
```
yanksyoon: I believe renaming these tests would require updating required status checks, should we discuss the tests design and perhaps finalise them?

CONTRIBUTING.md
```diff
@@ -93,6 +93,17 @@ Run unit tests using:
...
+Currently, the internal database package is only covered by integration tests. You can run the database integration tests using:
+
+```shell
+DB_CONNECT_STRING="postgres://postgres:password@localhost:5432/postgres?sslmode=disable" go test -cover -tags=integration -v  ./internal/database 
+```
+It assumes you have access to a Postgres server running reachable at $DB_CONNECT_STRING.
+You can use `docker` to run a Postgres server locally:
+```shell
+docker run -d --rm --name postgres -e POSTGRES_PASSWORD=password -p 5432:5432 postgres:latest
+```
```
yanksyoon: comment for myself: I think we can look into having a testing docker-compose since we'll soon be testing also with RABBITMQ & other components

yanksyoon (approved):
LGTM!

.github/workflows/tests.yaml
```diff
@@ -26,6 +26,21 @@ jobs:
           go fmt ./...
           git diff --exit-code
 
-      - name: Build and Test
+      - name: Run Unit and Integration Test of internal packages
+        env:
+          POSTGRESQL_CONNECT_STRING: postgres://postgres:postgres@localhost:5432/postgres
+
+        run: |
+          go test -v -coverprofile=coverage.out -race -tags=integration ./internal/...
```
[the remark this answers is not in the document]
yanksyoon: > we want to merge unit & internal integration tests for coverage right?

Oh it's stated in the PR body, sorry missed that. But my question on merging them still stands.

PR #36, opened 2025-11-19
feat: add get flavor pressure endpoint

florentianayuwono:
### Overview

Add functionality and test cases for the planner to serve `GET /api/v1/flavors/<str:name>/pressure` enpoint.

### Rationale

Part of Planner MVP.

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`, `documentation`).


internal/planner/telemetry.go
```diff
@@ -19,10 +19,7 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
-const (
-	flavorPlatform           = "github" // Currently only github is supported
-	flavorPressureMetricName = "mayfly_planner_flavor_pressure"
-)
+const flavorPressureMetricName = "mayfly_planner_flavor_pressure"
```
cbartz: Sorry we overlooked this. In previous PR's we agreed to use github_runner instead of mayfly. Look e.g. at the `telemetry` module in the webhook-gateway, the metrics are called like `github-runner.webhook.gateway.inbound.errors`

internal/database/database_test.go
```diff
@@ -671,6 +671,73 @@ func TestDatabase_GetPressures(t *testing.T) {
...
+		Labels:   []string{"self-hosted", "arm64", "large"},
+	}))
+
+	pressures, err := db.GetAllPressures(
+		ctx,
+		"github",
+	)
+	assert.NoError(t, err)
+	assert.Equal(t, map[string]int{
+		"github-amd64-medium-ps6": 1,
```
cbartz: Can you add one flavor without any job, so pressure should equal to 0 ?

cbartz (commented):
Thank you!

internal/planner/server_test.go
```diff
@@ -139,3 +142,65 @@ func TestCreateFlavor(t *testing.T) {
...
+			store := &fakeStore{errToReturn: tt.storeErr}
+			server := NewServer(store, NewMetrics(store))
+			token := makeToken()
+
+			req := newRequest(tt.method, tt.url, "", token)
+			w := httptest.NewRecorder()
+			server.ServeHTTP(w, req)
+
+			var resp map[string]int
+			json.NewDecoder(w.Body).Decode(&resp)
```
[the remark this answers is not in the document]
florentianayuwono: this is intentional as there are test cases where expected pressures is `nil` for negative cases.

yanksyoon (approved):
Minor nits, after @cbartz 's comment, i think this is already good to go, thanks!

internal/planner/server.go
```diff
@@ -92,3 +99,38 @@ func (s *Server) createFlavor(w http.ResponseWriter, r *http.Request) {
 
 	http.Error(w, fmt.Sprintf("failed to create flavor: %v", err), http.StatusInternalServerError)
 }
+
+// getFlavorPressure handles retrieving the pressure for a specific or all flavors.
+func (s *Server) getFlavorPressure(w http.ResponseWriter, r *http.Request) {
+	var (
+		pressures map[string]int
+		err       error
+	)
```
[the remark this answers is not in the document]
yanksyoon: Wouldn't it be also fine to just instantiate them within the switch statements like so `pressures, err := s.store.GetAllPressures(r.Context(), flavorPlatform)`?
The reason why I prefer the initialization when used or at least `map[string]int{}` type initialization is because maps can be tricky for initialization. This map in particular hasn't been initialized.
Helpful resource: https://medium.com/@mantianyu/golang-c153ff3838e2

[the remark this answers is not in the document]
yanksyoon: oh i see, thanks, I guess the best fix would be to initialize the map like so
```
var pressures = map[string]int{}
```
(the difference being `{}` at the end).

planner/main_test.go
```diff
@@ -22,107 +22,96 @@ import (
...
 
 	platform := "github"
 	labels := []string{"self-hosted", "amd64"}
 	priority := 42
 	flavor := randString(10)
+	pressure := 7
+
+	testCreateFlavor(t, port, flavor, platform, labels, priority, pressure)
+	testGetFlavorPressure(t, port, flavor, pressure)
+
```
cbartz: I would prefer more a test layout like

```
resp = createFlavor(...)
assertResponseIsHttpCreated(resp)
resp = getFlavor()
assertResponseIHttpOK(resp)
assertFlavorPressureEquals(resp, expected)
```

instead of the "sub tests" approach

cbartz (commented):
Thanks. Looks good, just some nitty comments on the test layout / comments.

internal/database/database.go
```diff
@@ -485,6 +485,40 @@ func (d *Database) GetPressures(ctx context.Context, platform string, flavors ..
 	return pressures, nil
 }
 
+// GetAllPressures returns the pressure for all flavors on the specified platform.
+func (d *Database) GetAllPressures(ctx context.Context, platform string) (map[string]int, error) {
```
florentianayuwono: see if we can modify the `GetPressures` function such that if there is no flavor provided, then we just fetch every flavors.

internal/planner/server.go
```diff
@@ -92,3 +99,36 @@ func (s *Server) createFlavor(w http.ResponseWriter, r *http.Request) {
...
+	default:
+		pressures, err = s.store.GetPressures(r.Context(), flavorPlatform, flavorName)
+	}
+
+	if err == nil {
+		respondWithJSON(w, http.StatusOK, pressures)
+		return
+	}
+
+	if errors.Is(err, database.ErrNotExist) {
```
florentianayuwono: right now the database won't return `ErrNotExist`, we can just replace this by checking if the pressures returned by the database is actually empty

internal/planner/server_test.go
```diff
@@ -30,13 +32,28 @@ func (f *fakeStore) AddFlavor(ctx context.Context, flavor *database.Flavor) erro
...
+	}
+	if len(res) == 0 {
+		return nil, database.ErrNotExist
 	}
 	return res, nil
 }
 
+func (f *fakeStore) GetAllPressures(ctx context.Context, platform string) (map[string]int, error) {
+	if f.pressures == nil {
+		return nil, database.ErrExist
```
florentianayuwono: same thing as above

florentianayuwono (commented):
review session from weii

cbartz (approved):
LGTM

PR #177, opened 2026-04-13
feat(logging): add tests for 5xx error logging and auth token leakage

yanksyoon:
### Overview

- Adds 5xx level error logging
- Adds changelog

### Rationale

- We are missing details on why 5xx errors happen in the planner logs.

### Checklist

- [x] Technical author has been assigned to review the PR in case of documentation changes (usually *.md files)

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,0 +1,7 @@
+# Changelog
+
+This changelog documents user-relevant changes to the Planner charm and Webhook gateway charm.
+
+## 2026-04-13
+
+- add 5xx error logging to planner routes.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -67,3 +67,5 @@ In this documentation
     How-to guides <how-to/index>
     Reference <reference/index>
     Explanation <explanation/index>
+    Changelog <changelog>
+    
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -67,5 +67,3 @@ In this documentation
     How-to guides <how-to/index>
     Reference <reference/index>
     Explanation <explanation/index>
-    Changelog <changelog>
-    
--- a/internal/planner/server.go
+++ b/internal/planner/server.go
@@ -145,7 +145,7 @@ func (s *Server) createFlavor(w http.ResponseWriter, r *http.Request) {
 	}
 
 	logger.ErrorContext(r.Context(), "failed to create flavor", "flavor", flavorName, "error", err)
-	http.Error(w, fmt.Sprintf("failed to create flavor: %v", err), http.StatusInternalServerError)
+	http.Error(w, "internal server error", http.StatusInternalServerError)
 }
 
 // listFlavors handles listing all flavors (admin-protected).
@@ -153,7 +153,7 @@ func (s *Server) listFlavors(w http.ResponseWriter, r *http.Request) {
 	flavors, err := s.store.ListFlavors(r.Context(), flavorPlatform)
 	if err != nil {
 		logger.ErrorContext(r.Context(), "cannot list flavors", "error", err)
-		http.Error(w, fmt.Sprintf("cannot list flavors: %v", err), http.StatusInternalServerError)
+		http.Error(w, "internal server error", http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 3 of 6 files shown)


docs/changelog.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+This changelog documents user-relevant changes to the Planner charm and Webhook gateway charm.
+
+## 2026-04-13
+
+- add 5xx error logging to planner routes.
```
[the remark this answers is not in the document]
yanksyoon: We want this as markdown.

docs/index.rst
```diff
@@ -67,3 +67,5 @@ In this documentation
     How-to guides <how-to/index>
     Reference <reference/index>
     Explanation <explanation/index>
+    Changelog <changelog>
```
florentianayuwono: i don't think changelog should be included in rtd? probably why copilot is commenting in changelong.md

internal/planner/server.go
```diff
@@ -238,6 +242,7 @@ func (s *Server) deleteFlavor(w http.ResponseWriter, r *http.Request) {
 	}
 	err := s.store.DeleteFlavor(r.Context(), flavorPlatform, flavorName)
 	if err != nil {
+		logger.ErrorContext(r.Context(), "cannot delete flavor", "flavor", flavorName, "error", err)
 		http.Error(w, fmt.Sprintf("cannot delete flavor: %v", err), http.StatusInternalServerError)
```
cbartz: unrelated to the PR so doesn't need to be addressed here: I think internal server errors should not give details in the http response body, as this could leak internals

internal/planner/server_logging_test.go
cbartz, line 1: claude says

```
  1. Missing coverage for auth token endpoints: The TestServer5xxLogsAtErrorLevel table doesn't include createAuthToken, listAuthTokens, or deleteAuthToken — even though logging was added to those handlers. Worth adding for completeness.          
  2. Missing coverage for updateJob: The updateJob handler has two new log statements (one for the initial ListJobs call, one for the UpdateJob call), but neither is tested.
  3. Missing coverage for getFlavorPressure: Three new log lines were added to this handler (subscribe error, get pressure error, streaming setup error). None are in the test table. 
```

internal/planner/server.go
```diff
@@ -144,13 +144,15 @@ func (s *Server) createFlavor(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	logger.ErrorContext(r.Context(), "failed to create flavor", "flavor", flavorName, "error", err)
```
cbartz, line 147: a different approach could be to add error logs in a logging middleware, so that we don't  forgot those in every handler.

cbartz (approved):
TY!

internal/planner/server_logging_test.go
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright 2025 Canonical Ltd.
```
florentianayuwono:
```suggestion
 * Copyright 2026 Canonical Ltd.
```

internal/planner/server_logging_test.go
```diff
@@ -0,0 +1,181 @@
...
+			url:    "/api/v1/jobs/github/42",
+			body:   `{"started_at":"2026-01-01T00:00:00Z"}`,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			withTestLogger(t, &buf, func() {
+				ticker := time.NewTicker(30 * time.Second)
```
florentianayuwono: why do we need a ticker in this test?

florentianayuwono (commented):
thanku :)

internal/planner/server.go
yanksyoon: I decided not to do so because a catch-all solution there may leak information that we don't want logged :)

internal/planner/server.go
yanksyoon: I'll scan through them and update them in this PR since the scope is pretty small :)

docs/index.rst
yanksyoon: Some projects like jenkins include it in rtd/ copilot is commenting because it is using .md rather than .rst in index file.

internal/planner/server_logging_test.go
yanksyoon: Good point, there's no requirement for heartbeath here, removed :)

florentianayuwono (approved):

PR #178, opened 2026-04-16
Disable logging for all juju deploy

yhaliaw:
### Overview

Disable logging of juju deploy

### Rationale

Prevent potential secret leaks.

### Checklist

- [ ] Changes comply with the project's coding standards and guidelines (see CONTRIBUTING.md and STYLE.md)

### DIFF
--- a/charms/tests/integration/conftest.py
+++ b/charms/tests/integration/conftest.py
@@ -123,7 +123,7 @@ def deploy_planner_app_fixture(
     resources = {
         "app-image": planner_app_image,
     }
-    juju.deploy(charm=planner_charm_file, app=app_name, resources=resources)
+    juju.deploy(charm=planner_charm_file, app=app_name, resources=resources, log=False)
     juju.wait(
         lambda status: jubilant.all_blocked(status, app_name),
         timeout=6 * 60,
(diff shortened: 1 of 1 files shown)


yanksyoon (approved):

weiiwang01 (approved):

charms/tests/integration/conftest.py
florentianayuwono, line 134: can we also disable logging for juju.config commands?

cbartz (approved):

charms/tests/integration/conftest.py
yhaliaw: I did think about that.

However, this is a newer charm, so the secrets would be passed in with juju secrets. Therefore, the `juju config` would not have secrets, and changing the config would be an important step in the tests. So I thought maybe keeping the logging would be fine.

There is definitely an argument to be made for disabling `juju.config` logging. For example, IP addresses are not considered secrets normally, but if the test is calling a resource internal to a company, the company might consider the IP address for an internal resource as private information.

Maybe we should see what the rest thinks.

yhaliaw:
The `log` argument does not exist for jubilant.

PR #185, opened 2026-04-22
feat(planner): resolve sub-30s queue time and surface median

cbartz:
### Overview

Two related improvements to the planner's "Job queue time" panel and the metric behind it:

1. **Bucket layout** (`internal/planner/telemetry.go`) — the waiting-time histogram no longer reuses the running-time bucket boundaries. Queue times are usually lower than job running times, so the new layout resolves the low end more finely (`1, 2.5, 5, 10, 15, 30s`) and caps the tail earlier (at 4h). Sub-30s queue times now land in distinct buckets instead of collapsing into a single one.

2. **Dashboard** (`charms/planner-operator/cos_custom/grafana_dashboards/go.json`) — the single "Job queue time" panel is split into two side-by-side panels: "Job queue time (median)" and "Job queue time (p95)". They live in the same dashboard slot the original panel occupied. The split is needed because the two series differ by orders of magnitude (median in seconds–minutes, p95 sometimes in hours), so a shared Y axis squashed the median to a flat line near zero. With separate panels each axis auto-scales independently. The panel descriptions also flag a caveat: `started_at − created_at` may overstate runner pickup time because the job's effective queued time can be later than `created_at`.

image

### Rationale

The previous p95-only view on coarse buckets was misleading in two ways: sub-30s waits showed up as ~28s due to linear interpolation inside the too-wide first bucket, and a single upper percentile without a central-tendency signal gave no sense of what a normal job experienced. Median + p95 on buckets tuned for queue times — and on separate auto-scaled panels — give a much more honest picture of the system's behaviour.

### Checklist

- [ ] `CONTRIBUTING.md` has been updated upon changes to the contribution/development process (e.g. changes to the way tests are run) — N/A
- [ ] Technical author has been assigned to review the PR in case of documentation changes (usually *.md files) — N/A

### DIFF
--- a/charms/planner-operator/cos_custom/grafana_dashboards/go.json
+++ b/charms/planner-operator/cos_custom/grafana_dashboards/go.json
@@ -536,7 +536,7 @@
         "type": "prometheus",
         "uid": "${prometheusds}"
       },
-      "description": "Time jobs waited between creation and starting, shown at the median (p50) and 95th percentile (p95)",
+      "description": "Median time jobs waited between creation and starting on a runner",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -594,7 +594,7 @@
       },
       "gridPos": {
         "h": 8,
-        "w": 12,
+        "w": 6,
         "x": 12,
         "y": 16
       },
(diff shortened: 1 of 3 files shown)


yhaliaw (approved):
Look good.
Adding the median would be a improvement.

florentianayuwono (approved):

PR #194, opened 2026-04-30
fix: use otlp exporter type instead of otlp_grpc

florentianayuwono:
## Problem

The `enable-log-forwarding` action was generating a collector config fragment using `otlp_grpc` as the exporter name/type. The `opentelemetry-collector` snap does not register a component type named `otlp_grpc`, causing the collector to fail to start immediately after the config was written:

```
'exporters' unknown type: "otlp_grpc" for id: "otlp_grpc"
(valid values: [nop otlp kafka prometheus zipkin debug otlphttp file prometheusremotewrite])
```

## Fix

Change `EXPORTER_NAME` from `"otlp_grpc"` to `"otlp/github_runner_optin"`. The `otlp` type uses gRPC by default — the `/github_runner_optin` suffix is just a distinguishing name. This matches the convention already used by the runner-injected `github.yaml` config (`otlp/mimir`).

## Smoke test improvements

- Remove hardcoded `otlp-endpoint: 127.0.0.1:4317` so the smoke test uses the real runner-injected endpoint and logs actually reach Loki
- Write test log entries after forwarding is configured so there is real data to verify in Loki


weiiwang01 (approved):

florentianayuwono:
fix tested to be working here https://github.com/canonical/platform-engineering-testing/pull/39

PR #213, opened 2026-05-26
Implement webhook redelivery

swetha1654:
#### What this PR does

Implement Webhook redelivery service

#### Checklist

- [x] Technical author has been assigned to review the PR in case of documentation changes (usually *.md files)


internal/redelivery/redelivery.go
```diff
@@ -0,0 +1,197 @@
...
+	}, nil
+}
+
+// NewDaemonWithClient creates a daemon with an injected client (for testing).
+func NewDaemonWithClient(cfg *Config, client GitHubClient) *Daemon {
+	return &Daemon{
+		config: cfg,
+		client: client,
+	}
+}
```
swetha1654: is this normal in Go?

charms/webhook-gateway-operator/charmcraft.yaml
```diff
@@ -21,6 +21,53 @@ extensions:
...
 config:
   options:
+    github-app-id:
+      type: int
+      optional: true
+      description: >-
+        The ID of the GitHub App to use for communication with GitHub.
+        If provided, the other github-app-* options must also be provided.
+        The GitHub App needs to have write permission for Webhooks.
+        Either this or github-token must be provided.
```
cbartz: worth mentioning that we can also use client id, which is the future where github is heading https://github.blog/changelog/2024-08-23-client-ids-are-now-included-in-app-api-responses/

client and app id should work interchangeable

charms/webhook-gateway-operator/charmcraft.yaml
```diff
@@ -21,6 +21,53 @@ extensions:
...
+        The path of the organisation or repository where the webhooks are registered. 
+        Format: <organisation> or <organisation>/<repository>.
+    github-token:
+      type: secret
+      optional: true
+      description: >-
+        The juju secret id of the token to use for communication with GitHub.
+        This can be a PAT with write admin:repo_hook or a fine-grained token
+        with write permission for Webhooks.
+        Either this or the GitHub App configuration must be provided.
```
cbartz: we don't need to support github token , this should also simplify the code

swetha1654: removed

charms/webhook-gateway-operator/charmcraft.yaml
swetha1654: Right now, `ghinstallation` is required to create a Transport for the Github client. and `ghinstallation.New` only accepts appID (int), not clientID (string) is not supported yet.

internal/redelivery/config.go
```diff
@@ -0,0 +1,84 @@
...
+	GitHubAppPrivateKeyEnvVar     = "APP_GITHUB_APP_PRIVATE_KEY_VALUE"
+	WebhookGitHubOrgEnvVar        = "APP_WEBHOOK_GITHUB_ORG"
+	WebhookGitHubRepoEnvVar       = "APP_WEBHOOK_GITHUB_REPO"
+	WebhookIDEnvVar               = "APP_WEBHOOK_ID"
+	RedeliveryIntervalEnvVar      = "APP_REDELIVERY_INTERVAL_SECONDS"
+)
+
+// ConfigFromEnv builds a redelivery config from environment variables.
+// Returns nil if redelivery is not configured (missing org or webhook ID).
+func ConfigFromEnv() *Config {
```
weiiwang01: For a parsing function, maybe we can change the return type to `(*Config, error)` so it can return an error on parsing failure instead of terminating the program?

cmd/webhook-gateway/main.go
```diff
@@ -59,5 +69,40 @@ func main() {
...
+
+func startRedeliveryDaemon(ctx context.Context, wg *sync.WaitGroup) {
+	cfg := redelivery.ConfigFromEnv()
+	if cfg == nil {
+		return
+	}
+	daemon, err := redelivery.NewDaemon(cfg)
+	if err != nil {
+		log.Fatalf("failed to create redelivery daemon: %v", err)
+	}
```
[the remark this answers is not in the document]
swetha1654: it should be fine for the charm to go in error state. Its easily reversible by resetting the charm configuration and running `juju resolve`.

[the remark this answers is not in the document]
swetha1654: discussed with @cbartz

charms/webhook-gateway-operator/charmcraft.yaml
swetha1654: conclusion: we use appID for now

internal/redelivery/redelivery.go
weiiwang01: If you use a `cfg` that can pass the validation in the constructor, you can actually swap the client after construction, for example, in a test:


```go
client := NewDaemon(testCfg)
client.client = testGitHubclient
```

internal/redelivery/redelivery.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	var redelivered int
+	for _, delivery := range deliveries {
+		if !isFailedRoutableDelivery(delivery) {
+			continue
+		}
+
+		err := d.client.RedeliverDelivery(ctx, d.config.GitHubOrg, d.config.GitHubRepo, d.config.WebhookID, delivery.GetID())
+		if err != nil {
```
weiiwang01, line 167: Maybe the `github-runner.webhook.redelivery.errors` counter should measure each individual webhook redelivery failure, instead of an entire batch failure? I mean, consider moving `redeliveryErrors.Add(ctx, 1)` from `runOnce` to here

.github/workflows/charms_integration.yaml
```diff
@@ -16,6 +16,9 @@ jobs:
       actions: read
     with:
       working-directory: .
+      test-secret-1-name: TEST_GITHUB_APP_ID
+      test-secret-2-name: TEST_GITHUB_APP_INSTALLATION_ID
+      test-secret-3-name: TEST_GITHUB_APP_PRIVATE_KEY
```
cbartz, line 21: Is this stored in bitwarden?

docs/changelog.md
```diff
@@ -35,6 +35,7 @@ Each revision is versioned by the date of the revision.
 
 ## 2026-06-08
 
+- Implement the webhook redelivery service.
```
cbartz: nit: this should be in a new changelog with a current date

charms/tests/integration/conftest.py
```diff
@@ -6,14 +6,19 @@
...
+from uuid import uuid4
 
 import jubilant
 import pytest
 import requests
+from tests.integration.helpers import create_github_app_client
 
 logger = logging.getLogger(__name__)
 
+GITHUB_PATH = "swetha1654/github-runner-operators"
```
cbartz: Can we avoid hardcoding this and passing via pytest conftest param?

charms/webhook-gateway-operator/charmcraft.yaml
```diff
@@ -23,6 +23,53 @@ extensions:
...
```
cbartz: it seems we don't support token auth here - we should remove that

charms/tests/integration/conftest.py
```diff
@@ -212,6 +217,26 @@ def integrate_webhook_gateway_rabbitmq_fixture(
...
+def github_test_hook_fixture():
+    """Create and cleanup a temporary webhook in the test repository."""
+    github_client = create_github_app_client()
+    repo = github_client.get_repo(GITHUB_PATH)
+    hook = repo.create_hook(
+        name="web",
+        events=["workflow_job"],
+        config={
+            # Unreachable URL is intentional; deliveries can fail and become redeliverable.
+            "url": f"http://192.168.0.1:8080/{uuid4().hex}",
```
cbartz: nit: can we use a self-explanatory constant for the URL

cbartz (commented):
Thank you! Looks good / some mostly minor comments

charms/tests/integration/conftest.py
swetha1654: I'll change it to be a github secret, as the other app values are also passed as secrets and its easier to handle all of them from the same place as they are related.

.github/workflows/charms_integration.yaml
swetha1654: @cbartz added them now under the note "GitHub App details for integration test in github-runner-operators" under the Canonical org inside Platform Engineering collection

cmd/webhook-gateway/main.go
```diff
@@ -59,5 +69,43 @@ func main() {
...
+		log.Fatalf("http server failed: %v", err)
+	}
+
+	wg.Wait()
+}
+
+func startRedeliveryDaemon(ctx context.Context, wg *sync.WaitGroup) {
+	cfg, err := redelivery.ConfigFromEnv()
+	if err != nil {
+		log.Fatalf("failed to parse redelivery config from environment: %v", err)
```
cbartz, line 97: This would put the charm in error state due to a config option issue - I think we should go in blocked state. 
To achieve this, we would need config validation in charm.py and raise a CharmConfigInvalidError, which would let paas-charm put the charm in blocked state.

cmd/webhook-gateway/main.go
```diff
@@ -59,5 +69,43 @@ func main() {
...
+	}
+	if cfg == nil {
+		return
+	}
+	daemon, err := redelivery.NewDaemon(cfg)
+	if err != nil {
+		log.Fatalf("failed to create redelivery daemon: %v", err)
+	}
```
cbartz, line 105: Same issue here. The new daemon creation could fail due to configuration.

CONTRIBUTING.md
```diff
@@ -209,6 +209,16 @@ You would need an LXD and MicroK8s cloud to run the tests. Ensure the `microk8s`
 controller is active in your Juju client before running the tests. An
 example run command in the root directory is as follows:
 
+Before running `webhook-gateway-integration`, export the GitHub App credentials
+required by `test_boostrap_webhook_redelivery.py`:
```
cbartz:
```suggestion
required by `test_webhook_redelivery.py`:
```

cmd/webhook-gateway/main.go
swetha1654: @cbartz I believe we discussed this, this would mean duplicating the validation logic on both the charm and the application side. The charm user can just reset the configuration and run `juju resolve` to get the charm back to active state and you mentioned that this is fine for now.
Would you still like me to add the validation logic on the charm side?

swetha1654: added it on the charm side anyway.

swetha1654: I haven't modified the go code. Ideally the charm should block before we hit validation errors in Go. If we do go into error state due to validation errors, it might be worth surfacing them as errors indicating the charm validation isn't sufficient.

charms/webhook-gateway-operator/src/charm.py
```diff
@@ -24,8 +39,45 @@ def __init__(self, *args: typing.Any) -> None:
         """
         super().__init__(*args)
 
+    def is_ready(self) -> bool:
```
cbartz, line 72: Couldn't we just  drop this , and raise CharmConfigInvalidError in _create_app function ?
This seems to be the more cleaner way with less code.

cc @javierdelapuente

charms/webhook-gateway-operator/src/charm.py
```diff
@@ -24,8 +39,45 @@ def __init__(self, *args: typing.Any) -> None:
...
+            return f"Incomplete redelivery config: missing {', '.join(missing)}"
+
+        invalid = [
+            f
+            for f in _REDELIVERY_INT_FIELDS
+            if (val := self.config.get(f)) is not None and (not isinstance(val, int) or val <= 0)
+        ]
+        if invalid:
+            return f"Invalid config (must be positive integers): {', '.join(invalid)}"
+
```
cbartz, line 60: There is one edge case - PEM validation of the private key. It is not validated here, though I guess this is fine (would be overkill). But if the PEM is wrong the charm will go in error.

cbartz (approved):
Looks good, some discussion about charm validation. Forward approving - assuming my comments are addressed.

charms/webhook-gateway-operator/src/charm.py
alithethird: I would argue `is_ready` is more clean since it is a dedicated method for checking if the app is ready. The `_create_app` function is just for instantiating the `App` class specific to your framework.

swetha1654: @cbartz Yeah i also iterated with copilot over the same question before i pushed the changes and it told me this which was quite convincing: 

> `is_ready()` is the framework's dedicated readiness gate. The paas_charm base class calls `is_ready()` before doing any work in `restart()` — if it returns False, the workload never starts. By overriding `is_ready()`, the validation plugs into the framework's existing flow rather than reimplementing the guard pattern. `is_ready()` answers "can this charm run?" — config validation is exactly that question.

charms/webhook-gateway-operator/src/charm.py
```diff
@@ -24,8 +39,45 @@ def __init__(self, *args: typing.Any) -> None:
...
+    def _validate_redelivery_config(self) -> str | None:
+        """Validate redelivery-related config options.
+
+        Returns:
+            An error message if validation fails, None otherwise.
+        """
+        present = [f for f in _REDELIVERY_REQUIRED_FIELDS if self.config.get(f)]
+        if not present:
+            return None
```
javierdelapuente: So no field is correct, but if any of them is set it goes to blocked? OOC, what is the reason for that?

javierdelapuente: Explained, this is not the only use of the charm.

charms/webhook-gateway-operator/src/charm.py
```diff
@@ -43,6 +95,24 @@ def gen_environment() -> dict[str, str]:
...
                 env["OTEL_TRACES_EXPORTER"] = "otlp"
                 env["OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"] = "http/protobuf"
+
+            if charm.config.get("github-path"):
+                github_path = str(charm.config["github-path"])
+                parts = github_path.split("/", 1)
+                env["APP_WEBHOOK_GITHUB_ORG"] = parts[0]
+                if len(parts) > 1:
+                    env["APP_WEBHOOK_GITHUB_REPO"] = parts[1]
+            if charm.config.get("webhook-id"):
```
javierdelapuente: This is not needed. By default all config options go as env vars in the workload.

charms/webhook-gateway-operator/src/charm.py
javierdelapuente: I do not have any preference for when to handle that (except as early as possible). I think `is_ready` is a good option.

Another alternative is to try to have a custom `framework_config_class`, which would give a pydantic validation (see https://github.com/canonical/kernelfactory.canonical.com/pull/525/changes#diff-00e1f5abacb18c6faa87ba0b1d57e0cdcaea302b04c0cf34dbe95342308ef0a3 for an example)

charms/webhook-gateway-operator/src/charm.py
swetha1654: There are other cases where the charm would go into error state, such as when the github authentication fails. For now the charm only does basic config validation.

charms/webhook-gateway-operator/src/charm.py
swetha1654: I like this approach better, let me use that. Its cleaner.

swetha1654: Due to [this](https://github.com/canonical/paas-charm/issues/301) issue in paas-charm, I cannot use a custom  `framework_config_class`. For now I use the `is_ready` override to validate the config and block the charm.

PR #234, opened 2026-06-15
feat(configurator): configurator integration

yanksyoon:
#### What this PR does

- Creates configurator integration
- Writes openstack configuration w/ clouds.yaml

#### Why we need it

- To have openstack provider ready

#### Checklist


charms/garm-configurator/src/charm.py
```diff
@@ -58,6 +65,46 @@ def _reconcile(self, event: ops.EventBase) -> None:
...
+                pc = state.provider_config
+                gc = state.github_app_config
+                sc = state.scaleset_config
+                garm_relation.data[self.unit].update(
+                    {
+                        "openstack_auth_url": pc.auth_url,
+                        "openstack_username": pc.username,
+                        "openstack_password": pc.password,
+                        "openstack_project_name": pc.project_name,
+                        "openstack_user_domain_name": pc.user_domain_name,
```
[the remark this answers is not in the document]
yanksyoon: Resolved — credentials are no longer written in plaintext to the relation databag.

The charm now creates a Juju secret for each sensitive value (password and private key), publishes the secret URI over the relation as ``openstack_password_secret_uri`` and ``github_private_key_secret_uri``, and grants the related application read access via ``secret.grant(relation)``.

The GARM charm was also updated to resolve these secrets via ``get_secret(id=uri)`` when reading relation data, so credentials never leave the Juju secret store.

charms/garm-configurator/src/charm.py
```diff
@@ -58,6 +65,46 @@ def _reconcile(self, event: ops.EventBase) -> None:
...
+                        "openstack_project_domain_name": pc.project_domain_name,
+                        "openstack_region_name": pc.region_name,
+                        "openstack_network": pc.network,
+                        "github_client_id": gc.client_id,
+                        "github_installation_id": gc.installation_id,
+                        "github_private_key": gc.private_key,
+                        "scaleset_name": sc.name,
```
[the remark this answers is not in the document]
yanksyoon: Resolved — see fix for comment #1 above. The GitHub private key is now stored in a Juju secret and only the secret URI is published over the relation as ``github_private_key_secret_uri``.

charms/tests/integration/conftest.py
```diff
@@ -505,3 +505,87 @@ def _on_image_relation_joined(self, event):
...
     )
     return app_name
+
+
+@pytest.fixture(scope="module", name="garm_configurator_charm_file")
+def garm_configurator_charm_file_fixture(pytestconfig: pytest.Config) -> str | None:
+    """Return the path to the built garm-configurator charm file."""
+    charm = pytestconfig.getoption(CHARM_FILE_PARAM)
+    if not charm:
+        return None
```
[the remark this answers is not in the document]
yanksyoon: Resolved — replaced the ``return None`` with ``pytest.skip()`` that mirrors the pattern already used in ``test_garm_configurator.py``. This produces a clear skip message and avoids passing ``None`` into ``juju.deploy()``.

charms/garm-configurator/src/charm.py
```diff
@@ -58,6 +65,46 @@ def _reconcile(self, event: ops.EventBase) -> None:
...
+                    }
+                )
+                if sc.repo is not None:
+                    garm_relation.data[self.unit]["scaleset_repo"] = sc.repo
+                if sc.org is not None:
+                    garm_relation.data[self.unit]["scaleset_org"] = sc.org
+                if sc.pre_install_scripts is not None:
+                    garm_relation.data[self.unit][
+                        "scaleset_pre_install_scripts"
+                    ] = sc.pre_install_scripts
```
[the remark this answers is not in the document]
yanksyoon: Resolved — before writing the updated relation data, the charm now explicitly removes stale optional keys (``scaleset_repo``, ``scaleset_org``, ``scaleset_pre_install_scripts``) that are no longer set in config. This prevents leftover values from coexisting with new config values.

charms/garm/src/charm.py
```diff
@@ -174,23 +314,49 @@ def restart(self, rerun_migrations: bool = False) -> None:
 
         # Short-circuit if postgresql relation data is not yet available.
         # GARM cannot start without a database connection.
-        if not self._get_postgresql_config():
+        postgresql_config = self._get_postgresql_config()
+        if not postgresql_config:
```
yanksyoon, line 344: This is guarded by paas lib - but is still done for typing in further code's sake

charms/garm/src/charm.py
```diff
@@ -174,23 +314,49 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        # (jwt_secret, db_passphrase, passwords).
+        provider_names = (
+            [p.get("unit_name") for p in provider_configs] if provider_configs else ["default"]
+        )
+        logger.info("Updating GARM config for providers: %s", provider_names)
+
+        container = self.unit.get_container(CONTAINER_NAME)
+        container.push(GARM_CONFIG_PATH, toml_content, make_dirs=True)
+        for path, content in provider_files.items():
+            container.push(path, content, make_dirs=True)
```
yanksyoon: garm config toml and clouds.yaml are created

charms/garm/charmcraft.yaml
```diff
@@ -28,4 +28,19 @@ requires:
...
+    interface: garm_configurator_v0
+    limit: 1
+    optional: true
+
+config:
+  options:
+    garm-listen-address:
+      type: string
+      default: "0.0.0.0"
+      description: Address GARM API server listens on.
```
yanksyoon: Perhaps this can be a hard-coded value. Wait for team's opinions.

yanksyoon (commented):
Review comments

yanksyoon (commented):
Review comments

charms/garm/src/charm.py
```diff
@@ -41,33 +45,164 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+        "    auth:",
+        f"      auth_url: {auth['auth_url']}",
+        f"      username: {auth['username']}",
+        f"      project_name: {auth['project_name']}",
+        f"      user_domain_name: {auth['user_domain_name']}",
+        f"      project_domain_name: {auth['project_domain_name']}",
+        f"      password: {auth['password']}",
+        f"    region_name: {region_name}",
+    ]
+    return "\n".join(lines) + "\n"
```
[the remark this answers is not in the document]
yanksyoon: Its more of an edge case, clouds yaml contents shouldn't have such chars. Keeping as current for simplicity.

charms/garm/charmcraft.yaml
```diff
@@ -28,4 +28,19 @@ requires:
...
+    garm-listen-port:
+      type: int
+      default: 8080
+      description: Port GARM API server listens on.
```
cbartz: I think we don't need those - we'll hard code the port (discussed with @javierdelapuente )

charms/garm/charmcraft.yaml
```diff
@@ -28,4 +28,19 @@ requires:
...
     optional: false
     limit: 1
+  garm-configurator:
+    interface: garm_configurator_v0
+    limit: 1
+    optional: true
+
+config:
+  options:
+    garm-listen-address:
```
cbartz: I think we don't need this config option. We will just bind on 0.0.0.0 . We are on k8s anyway, external access will go via ingress

charms/garm/src/charm.py
```diff
@@ -41,33 +45,164 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+                },
+            }
+        )
+    return result, provider_files
+
+
+def _render_clouds_yaml(cloud_name: str, auth: dict[str, str], region_name: str) -> str:
+    """Render a minimal clouds.yaml for gophercloud.
+
+    Since PyYAML is not a charm dependency, we render the YAML manually.
```
cbartz: why not using PyYAML ?  might be safer to use that for quoting issues etc

charms/garm/src/charm.py
```diff
@@ -41,33 +45,164 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+    lines = [
+        "clouds:",
+        f"  {cloud_name}:",
+        "    auth:",
+        f"      auth_url: {auth['auth_url']}",
+        f"      username: {auth['username']}",
+        f"      project_name: {auth['project_name']}",
+        f"      user_domain_name: {auth['user_domain_name']}",
+        f"      project_domain_name: {auth['project_domain_name']}",
+        f"      password: {auth['password']}",
```
cbartz: what happens if  'password' contains a char like :,# or similar that would break the yaml ?

charms/garm/src/charm.py
```diff
@@ -41,33 +45,164 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+        f"    region_name: {region_name}",
+    ]
+    return "\n".join(lines) + "\n"
+
+
 def render_garm_toml(
     *,
+    listen_address: str,
```
cbartz: I think we can hardcode that for now

charms/garm/src/charm.py
```diff
@@ -174,23 +314,54 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        if secrets_data is None:
+            logger.info("GARM secrets not yet available; blocking until leader initialises")
+            self.unit.status = ops.BlockedStatus("Waiting for GARM secrets")
+            return
+
+        # Render TOML including dynamic providers from Configurator relation
+        provider_configs = self._get_configurator_provider_configs()
+        toml_content, provider_files = render_garm_toml(
+            listen_address=str(self.config.get("garm-listen-address", "0.0.0.0")),
+            listen_port=int(self.config.get("garm-listen-port", GARM_PORT)),
```
cbartz: I think we don't need to support those config options

charms/garm/src/charm.py
```diff
@@ -205,6 +376,53 @@ def restart(self, rerun_migrations: bool = False) -> None:
             combine=True,
         )
         container.replan()
+        container.restart(PEBBLE_SERVICE_NAME)
```
cbartz: it seems the call is redundant as replan is already restarting all services ?

charms/garm/src/charm.py
```diff
@@ -174,23 +314,54 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+
+        hash_input = (
+            toml_content
+            + "\n"
+            + "\n".join(f"{path}\n{content}" for path, content in sorted(provider_files.items()))
+        )
+        new_hash = self._hash_toml(hash_input)
+        previous_hash = self._get_stored_toml_hash()
+        if previous_hash == new_hash:
+            logger.debug("TOML config unchanged; skipping restart")
```
cbartz, line 383: do we need to set the status to ACTIVE here ?

charms/garm/src/charm.py
```diff
@@ -265,30 +484,86 @@ def _get_postgresql_config(self) -> dict[str, typing.Any] | None:
...
+            password_secret_uri = data.get("openstack_password_secret_uri", "")
+            if not password and password_secret_uri:
+                password = self._resolve_secret_value(str(password_secret_uri))
+
+            # Resolve private key similarly.
+            private_key = data.get("github_private_key", "")
+            if not private_key:
+                pk_secret_uri = data.get("github_private_key_secret_uri")
+                if pk_secret_uri:
+                    private_key = self._resolve_secret_value(str(pk_secret_uri))
```
cbartz: nit: it seems the private key is not used in this PR , which is fine because it is not the scope of the PR. But the code could technically be removed.

charms/garm/src/charm.py
```diff
@@ -174,23 +314,54 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        if not postgresql_config:
             logger.info("PostgreSQL relation data not yet available; blocking")
             self.unit.status = ops.BlockedStatus("Waiting for postgresql relation")
             return
 
-        # TODO: Eliminate double-replan (ISD-5718). paas_charm calls replan()
-        # internally in super().restart(), which starts GARM with the default
-        # command momentarily before this method overrides it. Acceptable for
-        # the scaffold; resolve by contributing an upstream hook in a future story.
-        super().restart(rerun_migrations=rerun_migrations)
```
cbartz: the rerun_migrations super call is gone - don't we need that?

charms/garm/src/charm.py
cbartz, line 383: if this is run on a non-leader, we still might need to update the workload (because it checks the new hash against an application secret)

instead of comparing against an application secret, could we compare against the hash stored on the unit (using container.pull ) ?

cbartz (commented):
Thank you! I have some comments related to config options we dropped in the prometheus metrics PR (listen-port,listen-address).

charms/garm-configurator/charmcraft.yaml
```diff
@@ -114,6 +114,12 @@ requires:
     limit: 1
     optional: false
 
+provides:
+  garm-configurator:
+    interface: garm_configurator_v0
+    limit: 1
+    optional: true
```
florentianayuwono: just curious on why is it optional?

charms/garm/src/charm.py
florentianayuwono: another alternative is to use json n then dump it as yaml, at least with json no need to worry about spacing

charms/garm/charmcraft.yaml
yanksyoon: Done — removed `garm-listen-port` config option. Port is now hardcoded as `GARM_PORT = 8080` constant in `charm.py`.

charms/garm/charmcraft.yaml
yanksyoon: Done — removed `garm-listen-address` config option. Address is now hardcoded as `GARM_LISTEN_ADDRESS = "0.0.0.0"` constant in `charm.py`.

charms/garm/src/charm.py
yanksyoon: Done — `_render_clouds_yaml` now uses `yaml.dump()` from PyYAML. Added `PyYAML==6.0.3` to `requirements.txt`.

charms/garm/src/charm.py
yanksyoon: Fixed — `yaml.dump()` handles quoting automatically. Added a unit test (`test_render_clouds_yaml_quotes_special_chars`) that round-trips a password containing `:`, `#`, and `@` through YAML serialization to verify correct quoting.

charms/garm/src/charm.py
yanksyoon: Done — `listen_address` and `listen_port` are now module-level constants (`GARM_LISTEN_ADDRESS`, `GARM_PORT`) used directly in `render_garm_toml()`.

charms/garm/src/charm.py
yanksyoon: Done — removed both `garm-listen-address` and `garm-listen-port` from `charmcraft.yaml` config block, and removed the corresponding warning logic for `app-port`/`metrics-port`/`metrics-path` from `restart()`.

charms/garm/src/charm.py
yanksyoon: Removed — the entire `container.replan()` + `container.restart()` + `self.update_app_and_unit_status(ActiveStatus())` block is now replaced by `super().restart(rerun_migrations=...)`, which handles replan, ports, ingress, and ActiveStatus via the parent class.

charms/garm/src/charm.py
yanksyoon: The explicit `ActiveStatus` call is gone — `super().restart()` sets it on our behalf (the parent class calls `self.update_app_and_unit_status(ops.ActiveStatus())` at the end of its restart flow).

charms/garm/src/charm.py
yanksyoon: Done — removed the unused `private_key`/`github_private_key` resolution code from `_get_configurator_provider_configs()`.

charms/garm/src/charm.py
yanksyoon: Restored — `restart()` now calls `super().restart(rerun_migrations=rerun_migrations)` instead of manually doing `container.replan()` + `container.restart()`. The parent class handles migration re-runs, replan, ports, ingress, and ActiveStatus.

charms/garm/src/charm.py
yanksyoon: Done — replaced the app-level Juju secret hash (`_get_stored_toml_hash` / `_store_toml_hash`) with `_get_on_disk_toml_hash()` which reads the actual config files from the Pebble container and hashes them locally. This eliminates the race condition where non-leader units would read stale hashes due to app-secret propagation lag.

charms/garm/src/charm.py
```diff
@@ -174,23 +314,49 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        toml_content, provider_files = render_garm_toml(
+            listen_address=str(self.config.get("garm-listen-address", "0.0.0.0")),
+            listen_port=int(self.config.get("garm-listen-port", GARM_PORT)),
+            jwt_secret=secrets_data["jwt-secret"],
+            db_passphrase=secrets_data["db-passphrase"],
+            postgresql_config=postgresql_config,
+            providers=provider_configs if provider_configs else None,
+        )
+
+        new_hash = self._hash_toml(toml_content)
```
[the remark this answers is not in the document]
yanksyoon: Fixed — the hash input now includes both `toml_content` and all `provider_files` (sorted by path). The on-disk hash comparison also reads all provider files from the container, so credential rotations in provider configs will correctly trigger a restart.

charms/garm/src/charm.py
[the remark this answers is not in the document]
yanksyoon: Fixed — `_render_clouds_yaml` now uses `yaml.dump()` from PyYAML, which handles quoting automatically. Unit test `test_render_clouds_yaml_quotes_special_chars` verifies round-tripping of passwords containing `:`, `#`, and `@`.

charms/garm/src/charm.py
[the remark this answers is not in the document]
yanksyoon: Already addressed — all `container.push()` calls set `permissions=0o600` (owner read/write only). This was correct in the original code and is preserved.

charms/tests/integration/conftest.py
```diff
@@ -505,3 +505,115 @@ def _on_image_relation_joined(self, event):
         delay=10,
     )
     return app_name
+
+
+@pytest.fixture(scope="module", name="garm_configurator_charm_file")
+def garm_configurator_charm_file_fixture(pytestconfig: pytest.Config) -> str | None:
```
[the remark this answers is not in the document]
yanksyoon: This was already resolved in a previous iteration (`return None` replaced with `pytest.skip()`). The `str | None` annotation could be tightened to `str` since `None` is no longer returned, but this is a minor typing nit that doesn't affect behaviour.

charms/tests/integration/test_garm.py
```diff
@@ -443,3 +443,34 @@ def test_garm_metrics_endpoint_no_auth(
         "Expected the garm_health metric in the /metrics response; "
         f"got first 500 chars: {resp.text[:500]}"
     )
+
+
+def test_garm_toml_has_configurator_provider(
```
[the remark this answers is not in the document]
yanksyoon: The test name (`test_garm_api_has_configurator_provider`) accurately describes what it does — it verifies the GARM API serves the provider that was configured via the TOML. The name refers to the configuration mechanism, not the assertion method. Keeping as-is.

charms/garm-configurator/charmcraft.yaml
yanksyoon: It was initially marked `optional: true` to allow incremental rollout, but after discussion we've made it a required dependency. Both `charmcraft.yaml` files now have `optional` omitted (defaults to `false`), so the configurator relation must be present for the charm to reach ActiveStatus.

charms/garm/src/charm.py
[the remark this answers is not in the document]
yanksyoon: Fixed — the hash input now includes both `toml_content` and all `provider_files` (sorted by path). The on-disk hash comparison also reads all provider files from the container, so credential rotations in provider configs will correctly trigger a restart.

charms/garm/src/charm.py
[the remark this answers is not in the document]
yanksyoon: Fixed — `_render_clouds_yaml` now uses `yaml.dump()` from PyYAML, which handles quoting automatically. Unit test `test_render_clouds_yaml_quotes_special_chars` verifies round-tripping of passwords containing `:`, `#`, and `@`.

charms/garm/src/charm.py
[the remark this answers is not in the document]
yanksyoon: Already addressed — all `container.push()` calls set `permissions=0o600` (owner read/write only). This was correct in the original code and is preserved.

charms/tests/integration/conftest.py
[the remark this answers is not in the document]
yanksyoon: This was already resolved in a previous iteration (`return None` replaced with `pytest.skip()`). The `str | None` annotation could be tightened to `str` since `None` is no longer returned, but this is a minor typing nit that doesn't affect behaviour.

charms/tests/integration/test_garm.py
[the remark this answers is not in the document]
yanksyoon: The test name (`test_garm_api_has_configurator_provider`) accurately describes what it does — it verifies the GARM API serves the provider that was configured via the TOML. The name refers to the configuration mechanism, not the assertion method. Keeping as-is.

.github/workflows/charms_integration.yaml
```diff
@@ -44,6 +44,8 @@ jobs:
       rockcraft-channel: latest/edge
       charmcraft-channel: latest/edge
       test-tox-env: charms-integration
+      tmate-debug: true
+      tmate-timeout: 90
```
cbartz:
```suggestion
```

uv.lock
cbartz, line 1: do we need the lock here?

charms/garm/uv.lock
cbartz, line 1: do we need the lock here?

charms/garm-configurator/uv.lock
cbartz, line 1: do we need the lock here?

charms/garm/charmcraft.yaml
```diff
@@ -28,4 +28,7 @@ requires:
     interface: postgresql_client
     optional: false
     limit: 1
+  garm-configurator:
+    interface: garm_configurator_v0
+    limit: 1
```
cbartz, line 33: I think we should make the relation required. Without the relation, GARM will not really be functional. WDYT?

```suggestion
    optional: false
    limit: 1
```

charms/garm/src/charm.py
```diff
@@ -174,23 +305,53 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
-            logger.warning("garm-secrets not yet available; deferring config push")
-            self.unit.status = ops.WaitingStatus("Waiting for leader to initialise garm-secrets")
+        secrets_data = self._get_secrets()
+        if secrets_data is None:
+            logger.info("GARM secrets not yet available; blocking until leader initialises")
+            self.unit.status = ops.BlockedStatus("Waiting for GARM secrets")
             return
+
+        # Render TOML including dynamic providers from Configurator relation
+        provider_configs = self._get_configurator_provider_configs()
```
cbartz, line 357:
```suggestion
        provider_configs = self._get_configurator_provider_configs()
        if not provider_configs:
           self.unit.status = ops.WaitingStatus("Waiting for garm-configurator relation")
           return
```

charms/garm/src/charm.py
```diff
@@ -41,34 +46,155 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+
+    The garm-provider-openstack binary reads its credentials from a
+    provider-specific TOML config file that references a clouds.yaml file
+    -- it does NOT read OPENSTACK_* environment variables.
+
+    This function generates both the GARM provider entries and the
+    provider config file contents that must be pushed to the container.
+
+    If no providers are given, returns a default single-entry list with
+    the hardcoded "openstack" provider for backward compatibility.
```
cbartz: I think we should require a provider , so go into WaitingStatus earlier. This would also simplify the code here.

charms/garm/src/charm.py
```diff
@@ -174,23 +305,53 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
-        container = self.unit.get_container(CONTAINER_NAME)
-        try:
-            self._push_garm_config(container)
-        except ops.SecretNotFoundError:
-            logger.warning("garm-secrets not yet available; deferring config push")
-            self.unit.status = ops.WaitingStatus("Waiting for leader to initialise garm-secrets")
+        secrets_data = self._get_secrets()
+        if secrets_data is None:
+            logger.info("GARM secrets not yet available; blocking until leader initialises")
+            self.unit.status = ops.BlockedStatus("Waiting for GARM secrets")
```
cbartz:
```suggestion
            self.unit.status = ops.WaitingStatus("Waiting for GARM secrets")
```

cbartz (commented):
Thanks for the update.  I think the configurator relation should be mandatory for garm - and it should go in waiting if its not there. Without any configurator, the garm charm will not be really functional and an ACTIVE status could be misleading.

.github/workflows/charms_integration.yaml
cbartz: Mhh I still see the code

charms/garm/src/charm.py
```diff
@@ -203,46 +337,134 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        hash_input = (
+            toml_content
+            + "\n"
+            + "\n".join(
+                f"{path}\n{content}" for path, content in sorted(provider_files.items())
+            )
+        )
+        new_hash = self._hash_toml(hash_input)
+        previous_hash = self._get_on_disk_toml_hash(provider_files)
+        logger.info("hash: %s\ntoml: %s", new_hash, toml_content)
```
cbartz: this could leak secrets , we would need to escape the toml_content's secrets (jwt, db_passphrase)
```suggestion
```

charms/garm/src/charm.py
```diff
@@ -44,34 +48,157 @@ def _generate_passphrase(length: int = _DB_PASSPHRASE_LENGTH) -> str:
...
+
+    The garm-provider-openstack binary reads its credentials from a
+    provider-specific TOML config file that references a clouds.yaml file
+    -- it does NOT read OPENSTACK_* environment variables.
+
+    This function generates both the GARM provider entries and the
+    provider config file contents that must be pushed to the container.
+
+    If no providers are given, returns a default single-entry list with
+    the hardcoded "openstack" provider for backward compatibility.
```
cbartz: I think we can remove that sentence - there should also be not a case for backward compatiblity

charms/garm/src/charm.py
```diff
@@ -203,46 +337,134 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+        if not provider_configs:
+            self.unit.status = ops.WaitingStatus(
+                "Waiting for garm-configurator relation"
+            )
             return
+        toml_content, provider_files = render_garm_toml(
+            jwt_secret=secrets_data["jwt-secret"],
+            db_passphrase=secrets_data["db-passphrase"],
+            postgresql_config=postgresql_config,
+            providers=provider_configs if provider_configs else None,
```
cbartz: I think we require it to be not None

charms/garm/src/charm.py
```diff
@@ -203,46 +337,134 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
+            logger.debug("TOML config unchanged; skipping restart")
+            return
+
+        # Log non-sensitive metadata about the config change.
+        # Do NOT log toml_content here — it contains secrets
+        # (jwt_secret, db_passphrase, passwords).
+        provider_names = (
+            [p.get("unit_name") for p in provider_configs]
+            if provider_configs
+            else ["default"]
```
cbartz: I think this can be removed, we should assume there is at least one provider config

cbartz (commented):
Thank you! Mostly nits but one comment about secret leakage.

cbartz (approved):
Thank you!

PR #236, opened 2026-06-16
feat(garm): sync scalesets via GARM REST API without service restart (ISD-5729)

florentianayuwono:
#### What this PR does

Implements GARM scaleset sync via the GARM REST API without restarting the workload service (ISD-5729). When the `garm-configurator` relation data changes, the GARM charm diffs the desired scaleset state against the live REST API state and applies only the minimum set of POST/PUT/DELETE operations.

**Core implementation:**
- `garm_api.py` — thin wrapper around the generated OpenAPI client for scalesets, providers, credentials, and org/repo lookup
- `scaleset_reconciler.py` — `ScalesetReconciler` diffs desired vs observed scalesets and applies creates, updates, and deletes idempotently; defers creation when the referenced provider or credential is not yet registered (clean exit, no error state)
- `charm.py` — `_reconcile_scalesets()` is called from `restart()` on every config/relation change; `_on_update_status()` calls it directly (not via `restart()`) for periodic self-healing without triggering an unnecessary Pebble service restart; sets `WaitingStatus` on transient `GarmApiError` and `ActiveStatus` on success

**garm-configurator additions:**
- Writes scaleset spec fields (`provider_name`, `image`, `flavor`, `os_type`, `os_arch`, `min_idle_runners`, `max_runners`, `entity_type`, `entity_name`, `labels`, `pre_install_scripts`) to the `garm-configurator` relation databag
- New `os-type` config option (default: `linux`) propagated through to `ScalesetSpec` and `CreateScaleSetParams`

**Integration test infrastructure:**
- `fake-github-api` any-charm mock deployed alongside GARM; handles GitHub App installation tokens, runner groups, runner registration tokens, and scaleset CRUD endpoints
- Fixed `go-github v84` stripping `/api/v3/` prefix from enterprise URLs by normalising paths in both `do_GET` and `do_POST` handlers
- Fixed mock server IP detection in K8s using UDP socket trick and `Host` request header

#### Why we need it

Runner pool configuration must stay in sync with Configurator relation data without requiring a GARM service restart on every change. The reconciler diffs state and applies surgical API operations, so the workload is not disrupted.

#### Acceptance criteria met

- ✅ GARM's registered scalesets match the desired set derived from Configurator relation data
- ✅ Each scaleset reflects the current image, flavor, labels, `min_idle_runners`, `max_runners`, and `pre_install_scripts`
- ✅ Missing provider/credential defers scaleset creation and exits cleanly without error state
- ✅ Reconcile diffs observed REST state and applies the minimum set of POST/PUT/DELETE operations
- ✅ No service restart required for scaleset changes
- ✅ Integration tests verify create, update, delete, and deferred-creation scenarios

#### Checklist

- [ ] Changes comply with the project's coding standards and guidelines (see CONTRIBUTING.md and STYLE.md)
- [ ] **If this PR adds/removes a charm, or changes a charm's base class, conventions, or repo structure:** I updated the relevant `AGENTS.md`

### Test plan

- 265 GARM charm unit tests pass (unit + scaleset_reconciler + garm_api + generated client)
- 37 garm-configurator unit tests pass
- Integration test `test_scaleset_created_and_updated_via_relation` passes end-to-end: scaleset is created, updated, and deleted via relation data changes; deferred creation is verified when provider is absent

### Review focus

- `scaleset_reconciler.py`: `_needs_update()` diff logic and deferred-creation guard (`MissingDependencyError`)
- `charm.py` `_on_update_status`: calls `super()` first (preserves parent ingress refresh + DB migration logic) then `_reconcile_scalesets()` directly — avoids triggering a Pebble service restart on every periodic poll
- `conftest.py` fake GitHub API mock: `/api/v3` path stripping — required because `go-github v84` auto-prepends `/api/v3/` to enterprise base URLs
- `conftest.py` `configurator_garm` fixture: uses `all_agents_idle` for GARM (not `all_active`) because the unit workload status may be `waiting` while scaleset sync retries before first-run; `test_garm_charm_reaches_active` checks only `app_status.current` for the same reason

### Potential breaking changes

None — `os-type` defaults to `linux`; GARM charm falls back to `"linux"` for any existing relation data missing this field.


charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,141 @@
...
+    name: str
+    provider_name: str
+    credentials_name: str
+    image_id: str
+    flavor: str
+    os_arch: str
+    min_idle_runners: int
+    max_runners: int
+    labels: list[str] = field(default_factory=list)
+    runner_group: str = ""
```
florentianayuwono: im not sure what the default runner group should be

charms/garm/src/charm.py
```diff
@@ -414,6 +422,7 @@ def restart(self, rerun_migrations: bool = False) -> None:
         )
         container.replan()
         self._maybe_first_run()
+        self._reconcile_scalesets()
```
cbartz, line 443: bug: the reconcile scalesets code is never reached if the toml is unchanged due to  short-circuit logic in l394

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,159 @@
...
+    @staticmethod
+    def _needs_update(observed, spec: ScalesetSpec) -> bool:
+        observed_scripts = (observed.extra_specs or {}).get("pre_install_scripts", {})
+        return (
+            observed.image != spec.image
+            or observed.flavor != spec.flavor
+            or observed.max_runners != spec.max_runners
+            or observed.min_idle_runners != spec.min_idle_runners
+            or observed.github_runner_group != (spec.runner_group or None)
+            or observed_scripts != spec.pre_install_scripts
```
cbartz, line 219: should we also consider label difference ?

cbartz (commented):
Thank you! First pass. Need to look at tests still.

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,159 @@
...
+            if spec.name in observed:
+                self._maybe_update(observed[spec.name], spec)
+            else:
+                self._create(spec, entity_id)
+
+        for name, scaleset in observed.items():
+            if name not in all_desired_names:
+                logger.info("Deleting orphaned scaleset %s (id=%s)", name, scaleset.id)
+                if scaleset.id is not None:
+                    self._client.delete_scaleset(scaleset.id)
```
cbartz: There is an issue here. It seems GARM returns a 400 if the scaleset still contains runners. This would be logged and then ignored.

We would need to ensure that the scaleset has zero runners first using update_scaleset (+enabled=False)

Afterwards we can do a delete scaleset. There might be a 400 because the runners have not been cleaned up yet. IN this case the scaleset will be cleaned up upon next reconcile.

We could also check using `GET /scalesets/{scalesetID}/instances,` the number of instances for a scaleset to avoid a 400.

charms/garm/src/scaleset_reconciler.py
cbartz: I think Github uses `Default` as name for the Default Group

cbartz (commented):
second pass

charms/garm/tests/unit/test_garm_api.py
cbartz, line 1: nit: Test docstring should follow arrange/act/assert pattern

charms/garm/tests/unit/test_scaleset_reconciler.py
cbartz, line 1: nit: test docstrings should follow arrange/act/assert pattern

charms/tests/integration/test_garm.py
cbartz, line 1: nit: same here please follow arrange/act/assert test doc string pattern

charms/tests/integration/test_garm.py
```diff
@@ -474,30 +762,73 @@ def test_garm_metrics_endpoint_no_auth(
...
+        pytest.skip(
+            f"requires GARM provider {relation_data['provider_name']!r}; "
+            f"available: {sorted(provider_names)}"
+        )
+
+    _set_scaleset_relation_data(
+        juju,
+        garm_app,
+        garm_configurator_for_scaleset_tests,
+        min_idle_runner="1",
```
cbartz:
```suggestion
        min_idle_runner="0",
```

we should set it to zero here - we don't want a test that tries to spawn runners here. There will be an end to end test later as discussed.

charms/tests/integration/test_garm.py
cbartz, line 1: per contribution guide, we follow the "step down rule", so the callee functions should appear below the callers https://github.com/canonical/platform-engineering-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-ordering

cbartz (commented):
another pass

charms/garm-configurator/src/charm.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, *args: typing.Any) -> None:
         for event in [
             self.on.config_changed,
             self.on.secret_changed,
+            self.on[GARM_RELATION_NAME].relation_changed,
             self.on[IMAGE_RELATION_NAME].relation_joined,
```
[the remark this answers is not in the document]
florentianayuwono: fixed

charms/garm-configurator/src/charm.py
```diff
@@ -4,6 +4,7 @@
 
 """Charm entrypoint for the GARM configurator charm."""
 
+import json
 import typing
```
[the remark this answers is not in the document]
florentianayuwono: the pre install script is not meant to be like dict-like, therefore the current implementation is fine

charms/garm/src/garm_api.py
```diff
@@ -140,3 +156,272 @@ def first_run(
...
+            try:
+                return (
+                    ProvidersApi(api_client=client).list_providers(
+                        _request_timeout=_REQUEST_TIMEOUT
+                    )
+                    or []
+                )
+            except ApiException as exc:
+                raise GarmApiError(f"Failed to list providers ({exc.status}): {exc.body}") from exc
+
```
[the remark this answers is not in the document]
florentianayuwono: fixed

charms/garm/src/charm.py
florentianayuwono: i have added the reconcile call to l394 before the return statement as well

charms/garm/src/scaleset_reconciler.py
florentianayuwono: label difference is checked now, although it's not supported to be updated (only delete and recreate can be done) so for now warning is issued

charms/tests/integration/test_garm.py
florentianayuwono: file refactored!

charms/garm-configurator/src/charm.py
```diff
@@ -109,17 +122,57 @@ def _ensure_relation_secret(
...
+            "name": state.scaleset_config.name,
+            "provider_name": state.provider_config.provider_name,
+            "image_id": state.image_id or "",
+            "flavor": state.scaleset_config.flavor,
+            "os_arch": state.scaleset_config.os_arch,
+            "min_idle_runner": str(state.scaleset_config.min_idle_runner),
+            "max_runner": str(state.scaleset_config.max_runner),
+            "labels": (
+                state.scaleset_config.labels
+                if isinstance(state.scaleset_config.labels, str)
```
yanksyoon: Is the `ininstance` check for string or list be required? Would the validation be performed in the state already?

charms/garm/src/charm.py
```diff
@@ -373,14 +385,13 @@ def restart(self, rerun_migrations: bool = False) -> None:
...
-                f"{path}\n{content}" for path, content in sorted(provider_files.items())
-            )
+            + "\n".join(f"{path}\n{content}" for path, content in sorted(provider_files.items()))
         )
         new_hash = self._hash_toml(hash_input)
         previous_hash = self._get_on_disk_toml_hash(provider_files)
         if previous_hash == new_hash:
-            logger.debug("TOML config unchanged; skipping restart")
+            logger.debug("TOML config unchanged; skipping service restart")
+            self._reconcile_scalesets()
```
yanksyoon, line 411: It looks like scaleset reconciliation depends rather on the service availability than any configuration changes - should the code be refactored to reflect that? (perhaps as a follow up). The current code paths seem to introduce a diverging paths for reconciliation.

charms/garm/src/charm.py
```diff
@@ -664,6 +670,89 @@ def _maybe_first_run(self) -> None:
...
+                            for label in data.get("labels", "").split(",")
+                            if label.strip()
+                        ],
+                        runner_group=data.get("runner_group", ""),
+                        pre_install_scripts=_parse_pre_install_scripts(
+                            data.get("pre_install_scripts", "")
+                        ),
+                    )
+                )
+        return specs
```
yanksyoon, line 755: As a follow up for the team, I think we should refactor relation/configuration parsing into the charm state since this file is getting pretty large.

charms/garm/src/charm.py
```diff
@@ -664,6 +670,89 @@ def _maybe_first_run(self) -> None:
...
+    def _reconcile_scalesets(self) -> None:
+        """Sync GARM scalesets against garm-configurator relation data."""
+        admin_creds = self._get_admin_credentials()
+        if not admin_creds:
+            logger.warning("Admin credentials not yet available; deferring scaleset reconcile")
+            return
+
+        garm_url = f"http://127.0.0.1:{GARM_PORT}"
+        try:
+            garm_client = GarmApiClient(f"{garm_url}/api/v1")
```
yanksyoon: As a follow up for the team, we should think about refactoring the GarmApiClient to be initialized and passed around beforehand since multiple instances are initialized throughout a single reconciliation.

charms/garm/src/garm_api.py
```diff
@@ -140,3 +156,288 @@ def first_run(
...
+            except ApiException as exc:
+                raise GarmApiError(f"GARM login failed ({exc.status}): {exc.body}") from exc
+            except urllib3.exceptions.HTTPError as exc:
+                raise GarmConnectionError(f"GARM connection error: {exc}") from exc
+        if not result.token:
+            raise GarmApiError("login response did not contain a token")
+        return result.token
+
+
+class GarmAuthenticatedClient:
```
yanksyoon: Should this class derive the bass GarmApiClient? It seems like an overkill to have two separate class instances to interact with GARM. WDYT?

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+        Performs the minimum set of CREATE / UPDATE / DELETE operations.
+        If a referenced provider is missing or the target entity (org/repo)
+        is not registered in GARM, that spec is skipped silently (deferred
+        creation) — no error state is set.
+
+        Args:
+            desired: The full desired set of scalesets.
+        """
+        providers = {p.name for p in self._client.list_providers()}
+        observed = {ss.name: ss for ss in self._client.list_scalesets()}
```
yanksyoon:
```suggestion
        providers = {provider.name for provider in self._client.list_providers()}
        observed = {scaleset.name: scaleset for scaleset in self._client.list_scalesets()}
```
As per our contribution guide, we prefer names that are not abbreviated for clarity :D

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+                "image": spec.image,
+                "flavor": spec.flavor,
+                "os_arch": spec.os_arch,
+                "min_idle_runners": spec.min_idle_runners,
+                "max_runners": spec.max_runners,
+                "labels": sorted(spec.labels),
+                "github_runner_group": spec.runner_group or None,
+                "extra_specs": extra_specs or None,
+            }
+        )
```
yanksyoon: Would validating the model here be a good idea? Perhaps the validation of the scaleset specs should be done earlier in the control loop (where it exits early) and then we can start using trusted the values afterwards.

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+            logger.debug("Scaleset %s is up to date", spec.name)
+            return
+
+        observed_labels = sorted(t.name for t in (observed.tags or []) if t.name)
+        if observed_labels != sorted(spec.labels):
+            # UpdateScaleSetParams has no labels field; label changes require
+            # recreating the scaleset. Log a warning so operators are aware.
+            logger.warning(
+                "Scaleset %s labels changed (%s -> %s) but cannot be updated in place;"
+                " delete and recreate the scaleset to apply label changes",
```
yanksyoon: I guess something to discuss with the team, do we want this to delete and recreate the scaleset automatically? Should this be an operator's duty to remove the scaleset and redefine them for delete & create?

charms/garm/tests/unit/test_scaleset_reconciler.py
```diff
@@ -0,0 +1,310 @@
...
+"""Unit tests for the scaleset reconciler."""
+
+import pytest
+
+from scaleset_reconciler import ScalesetReconciler, ScalesetSpec
+
+
+# ---------------------------------------------------------------------------
+# Fake client — records operations as observable state
+# ---------------------------------------------------------------------------
```
yanksyoon: I think we can remove these comment artifacts :D

yanksyoon (commented):
Some nitpicky comments, looks good, thank you @florentianayuwono !

charms/garm/src/scaleset_reconciler.py
cbartz: I think it should be the charms duty. I mean, charms are all about zero-ops and this looks like a valid use-case (we are doing this from time to time as well). 

To unblock this PR, I am happy to create a follow-up ticket (because I think this might need a bit of discussion how to do it).

charms/garm/tests/unit/test_scaleset_reconciler.py
cbartz: Yes I think the LLM's are repeating existing patterns here. Somehow it got introduced in the codebase.

charms/garm/src/garm_api.py
```diff
@@ -140,3 +156,288 @@ def first_run(
...
+                    or []
+                )
+            except ApiException as exc:
+                raise GarmApiError(
+                    f"Failed to list repositories ({exc.status}): {exc.body}"
+                ) from exc
+            except urllib3.exceptions.HTTPError as exc:
+                raise GarmConnectionError(f"GARM connection error: {exc}") from exc
+        for repo in repos:
+            if repo.name == repo_name:
```
cbartz: bug: repo_name has org/name format but `repo.name` is only the name

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+    provider_name: str
+    image: str
+    flavor: str
+    os_arch: str
+    min_idle_runners: int
+    max_runners: int
+    entity_type: str
+    entity_name: str
+    labels: list[str] = field(default_factory=list)
+    runner_group: str = "Default"
```
cbartz, line 33: can you set the default in the garm-configurator charm to uppercase "Default" as well?

charms/garm/src/garm_api.py
```diff
@@ -140,3 +156,288 @@ def first_run(
...
+                        _request_timeout=_REQUEST_TIMEOUT
+                    )
+                    or []
+                )
+            except ApiException as exc:
+                raise GarmApiError(f"Failed to list providers ({exc.status}): {exc.body}") from exc
+            except urllib3.exceptions.HTTPError as exc:
+                raise GarmConnectionError(f"GARM connection error: {exc}") from exc
+
+    def list_credentials(self) -> list[ForgeCredentials]:
```
cbartz: this seems to be not called - can we remove it please

charms/garm/src/charm.py
```diff
@@ -664,6 +670,89 @@ def _maybe_first_run(self) -> None:
...
+            token = garm_client.login(admin_creds["username"], admin_creds["password"])
+            auth_client = GarmAuthenticatedClient(f"{garm_url}/api/v1", token)
+            desired = self._build_desired_scalesets()
+            ScalesetReconciler(auth_client).reconcile(desired)
+        except GarmApiError as exc:
+            logger.warning("GARM API error during scaleset reconcile: %s", exc)
```
cbartz, line 775: A transient error would mean we would need to wait until the next event to have the desired scaleset and the charm stays active/idle, giving the operator the impression that there are no issues.

I guess the recommended pattern would be to use error state and let juju retry , but error state is something we would like to avoid (cc @javierdelapuente ).

I see two options

- add retry (similar as done for _maybe_status)
- add a "BlockedStatus" or something indicating the operator something went wrong

charms/garm/src/scaleset_reconciler.py
cbartz: I am happy to remove it for now in this PR.
We can consider the label case as follow-up to unblock this PR as this might need some discussion.

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+                except GarmApiError as exc:
+                    logger.warning("Could not disable scaleset %s before delete: %s", name, exc)
+                try:
+                    self._client.delete_scaleset(scaleset.id)
+                except GarmApiError as exc:
+                    # 400 means runners are still present; scaleset will be deleted
+                    # on the next reconcile pass once GARM has cleaned them up.
+                    logger.warning(
+                        "Could not delete scaleset %s (runners may still be active; "
+                        "will retry on next reconcile): %s",
```
cbartz: I think we should observe update_status as well, otherwise the scaleset might not be cleaned up for a long time (if there are no events fired). Though we need to make sure we also respect the parent class (paas-charm) update_status (which updates ingress,etc).

Something like

```
@block_if_invalid_data
def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
    """Run the base update-status handling, then reconcile so stuck deletes
    and transient GARM API errors self-heal on the periodic poll."""
    super()._on_update_status(event)
    self.restart()
```

@javierdelapuente any advice here?

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,191 @@
...
+        for name, scaleset in observed.items():
+            if name not in all_desired_names:
+                logger.info("Deleting orphaned scaleset %s (id=%s)", name, scaleset.id)
+                if scaleset.id is None:
+                    continue
+                try:
+                    # Disable the scaleset first so GARM stops launching new runners.
+                    # GARM returns 400 if the scaleset still has active runners,
+                    # so disabling first drains it for the next reconcile to clean up.
+                    self._client.update_scaleset(scaleset.id, UpdateScaleSetParams(enabled=False))
```
cbartz:
```suggestion
                    self._client.update_scaleset(scaleset.id, UpdateScaleSetParams(enabled=False, min_idle_runners=0))
```

otherwise the runners are kept and scaleset cannot be deleted

cbartz (commented):
Thank you. Some follow-up comments.

charms/garm/src/charm.py
florentianayuwono: i agree, we shld do it as follow up refactor

charms/garm/src/garm_api.py
florentianayuwono: i think good of your suggestion :) we can now use GarmApiClient for use cases when we don't have token yet (checking if garm is up, first run, login) and GarmAuthenticatedClient for everything else (scaleset crud, org repo lookups, provider listing) in addition

charms/garm/src/charm.py
florentianayuwono: i opted to have waiting status instead, because if we go into error state, it will need juju resolve to resolve, meanwhile we have update_status that will periodically rerun the reconcile scaleset, which will naturally retry reconciling and set the unit status back to active when everything is well.

charms/garm-configurator/charmcraft.yaml
```diff
@@ -77,6 +77,11 @@ config:
       type: string
       description: |
         The CPU architecture for runners in the scaleset.
+    os-type:
+      type: string
+      default: linux
+      description: |
+        The operating system type for runners in the scaleset (linux or windows).
```
cbartz: I think we don't need this, we'll only support Linux for now.

charms/garm/src/charm.py
cbartz: Agree but WaitingStatus is the wrong unit status here. This is usually used for relations to come in.
I am fine having update status for general reconcile (e.g. deletion of scaleset) but for a generic GarmApiError we should error out.

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,220 @@
...
+            logger.debug("Scaleset %s is up to date", spec.name)
+            return
+
+        observed_labels = sorted(t.name for t in (observed.tags or []) if t.name)
+        if observed_labels != sorted(spec.labels):
+            # UpdateScaleSetParams has no labels field; label changes require
+            # recreating the scaleset. Log a warning so operators are aware.
+            logger.warning(
+                "Scaleset %s labels changed (%s -> %s) but cannot be updated in place;"
+                " delete and recreate the scaleset to apply label changes",
```
cbartz: How should the operator delete? I assume via removing the configurator relation for that scaleset. We should add this information to be as explicit as possible.

charms/garm/src/charm.py
```diff
@@ -249,6 +251,19 @@ def _generate_admin_password() -> str:
...
 
+def _parse_pre_install_scripts(raw: str) -> dict[str, str]:
+    """Parse pre_install_scripts from JSON relation data string."""
+    if not raw:
+        return {}
+    try:
+        result = json.loads(raw)
+        if isinstance(result, dict):
+            return result
+    except (ValueError, json.JSONDecodeError):
```
cbartz: JSONDecodeError is a subclass of ValueError

```suggestion
    except ValueError:
```

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _create(self, spec: ScalesetSpec, entity_id: str, params: CreateScaleSetParams) -> None:
+        logger.info("Creating scaleset %s under %s %s", spec.name, spec.entity_type, entity_id)
+        if spec.entity_type == "organization":
+            self._client.create_org_scaleset(entity_id, params)
+        else:
+            self._client.create_repo_scaleset(entity_id, params)
+
+    def _maybe_update(self, observed, spec: ScalesetSpec) -> None:
+        if not self._needs_update(observed, spec):
```
cbartz: lets move the order - put needs_update below the label diff check , and then remove the label diff in needs_update . 

labeling is not leading to an update and we can avoid the redundant PATCH call in case only labels are changed

charms/garm/src/scaleset_reconciler.py
```diff
@@ -0,0 +1,220 @@
...
+        logger.info("Updating scaleset %s (id=%s)", spec.name, observed.id)
+        if observed.id is None:
+            logger.warning("Scaleset %s has no id; skipping update", spec.name)
+            return
+        self._client.update_scaleset(observed.id, params)
+
+    @staticmethod
+    def _needs_update(observed, spec: ScalesetSpec) -> bool:
+        observed_scripts = (observed.extra_specs or {}).get("pre_install_scripts", {})
+        observed_labels = sorted(t.name for t in (observed.tags or []) if t.name)
```
cbartz: lets drop the label diff in needs_update , aligned with my comment above

charms/garm/src/charm.py
florentianayuwono: can we still keep the waiting status for now? im a bit worried with erroring out as it will affect the current integration test (that's alr p complex). the flow currently is when garm-configurator is integrated w garm, the `_reconcile_scalesets()` will run before the garm's api is ready (race condition w first-run), so it will raise an exception. if we let the unit goes to error now, the configurator_garm fixture will be affected as it waits for the `all_agents_idle` and all subsequent hooks also will be blocked.

charms/garm-configurator/src/charm_state.py
cbartz: Can't we remove the whole plumbing and just hardcode this in GARM? I think non-linux is really out of scope for now.

cbartz (approved):
Forward-approving.

PR #242, opened 2026-06-18
feat(garm): route workload traffic through Juju model proxy

cbartz:
#### What this PR does

Makes the GARM charm honour the Juju model proxy (`juju-http-proxy` / `juju-https-proxy` / `juju-no-proxy`), which the charm previously had no way to configure. The values are applied to the GARM workload so both GARM's GitHub egress and the OpenStack provider executable (`garm-provider-openstack`) traverse the proxy.

#### How it works

- `_proxy_environment()` reads the `JUJU_CHARM_*_PROXY` hook variables and sets each non-empty value (lower- and upper-case) on the Pebble service environment.
- The proxy variable *names* are whitelisted in each provider's `external.environment_variables`, so GARM forwards them to the provider executable (external providers don't inherit GARM's environment). Those names are part of the rendered TOML, so enabling/disabling the proxy changes the config hash and triggers a replan.
- Known limitation (documented in code): changing a proxy *value* while the same variables stay set isn't detected until the next config change — to be addressed when GARM startup moves to an env-driven wrapper script.

#### Test plan

- Unit tests: `_proxy_environment` happy-path + edge cases, provider `environment_variables` present/absent, and proxy vars injected into the Pebble layer. Full garm unit suite passes (114); ruff/format/pyright/codespell clean.

#### Breaking changes

None — behaviour is byte-identical when no proxy is configured on the model.

#### Checklist


charms/garm/src/charm.py
```diff
@@ -375,11 +420,19 @@ def restart(self, rerun_migrations: bool = False) -> None:
             self.unit.status = ops.WaitingStatus("Waiting for garm-configurator relation")
             return
 
+        proxy_env = _proxy_environment()
+        # The proxy variable *names* are rendered into the TOML (each provider's
+        # environment_variables allowlist), so toggling the proxy on or off changes the
+        # config hash below and triggers a restart. The proxy *values* live only in the
+        # Pebble service environment; changing a value without changing which variables are
```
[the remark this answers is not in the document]
cbartz: This is a known, deliberate limitation — it's documented in the comment right above this block. The common cases (enabling or disabling the proxy) *are* detected, because that changes which variables are set and therefore the rendered `environment_variables` allowlist and the hash; only changing a proxy *value* while the same variables stay set is missed, and it self-corrects on the next config change.

We intentionally chose this over the two alternatives you mention:
- Folding the proxy *values* into the hash would compare them against the on-disk TOML hash (where the values don't exist), so the hashes would never match and the charm would push config + replan on *every* hook.
- Independently detecting service-env drift means reading the applied values back from the Pebble plan — an earlier revision of this PR did exactly that, and we removed it to keep the change small.

Proper value-change detection will come for free when GARM startup moves to the env-driven wrapper script (generate the TOML from env, derive secrets via paas-charm), where the proxy values become part of the tracked state. Tracked as a follow-up; keeping this PR minimal.

_AI-generated message, on behalf of the PR author._

charms/garm/src/charm.py
```diff
@@ -133,15 +169,19 @@ def _build_provider_list(
         provider_files[provider_toml_path] = provider_toml
         provider_files[clouds_yaml_path] = clouds_yaml
 
+        external: dict[str, typing.Any] = {
+            "config_file": provider_toml_path,
+            "provider_executable": OPENSTACK_PROVIDER_BINARY,
+        }
+        if proxy_var_names:
+            external["environment_variables"] = proxy_var_names
+
```
florentianayuwono: this function is getting quite long, im wondering if it's better to refactor this into helper function, since there's  some code duplication as well

florentianayuwono (approved):
lgtm thanks chris!

charms/garm/src/charm.py
cbartz: Good call — extracted a `_build_external_block(config_file, proxy_var_names)` helper in 568af83 that both the default and per-provider branches now use, removing the duplicated `external` dict construction. Unit suite still green (114 passing).

---
*AI-generated message.*

cbartz: addressed

PR #251, opened 2026-06-23
Feat/tmate integration

yhaliaw:
#### What this PR does

Adding the `debug-ssh` relation support for tmate-server integration with GARM charm.

##### Summary
This pull request introduces support for injecting SSH debug (tmate) connection information into GARM runner install templates via a new `debug-ssh` integration. It adds a new relation, parses and validates SSH debug data, and ensures that the runner install template is dynamically created or updated to include this data. Additionally, it extends the GARM API client to support template management operations. 

The most important changes are:

**Debug SSH Integration & State Management**
* Added a new `debug-ssh` relation to `charmcraft.yaml` and registered event handlers in `charm.py` to trigger reconciliation when debug-ssh relation events occur. [[1]](diffhunk://#diff-017551dbc9e8a62f2cc4e05e76ce9f85f81050b92d4ab40bd258625b5e75ef8fR35-R37) [[2]](diffhunk://#diff-37f5e1d5b8f0cfc70d209397d70fd52d7bb6d2033a5e8debe5ba96b4440921d8R303-R314)
* Introduced a new `charm_state.py` module to parse and validate SSH debug connection info from the `debug-ssh` relation, exposing it as immutable state for use during reconciliation.

**Runner Template Injection**
* Updated the runner reconciliation logic to ensure the `github_linux_charmed` template is present and up-to-date, injecting SSH debug environment variables into the runner install script if debug-ssh connections are available. [[1]](diffhunk://#diff-37f5e1d5b8f0cfc70d209397d70fd52d7bb6d2033a5e8debe5ba96b4440921d8L802-R822) [[2]](diffhunk://#diff-37f5e1d5b8f0cfc70d209397d70fd52d7bb6d2033a5e8debe5ba96b4440921d8L821-R844)
* Modified the `ScalesetSpec` builder to accept and propagate the template ID for use by scalesets. [[1]](diffhunk://#diff-37f5e1d5b8f0cfc70d209397d70fd52d7bb6d2033a5e8debe5ba96b4440921d8L675-R689) [[2]](diffhunk://#diff-37f5e1d5b8f0cfc70d209397d70fd52d7bb6d2033a5e8debe5ba96b4440921d8R752)

**GARM API Client Extensions**
* Extended the GARM API client with methods to list, get, create, update, and delete runner install templates, and added utility functions for injecting tmate environment snippets into shell scripts. [[1]](diffhunk://#diff-f632e2877b1f332884af3c5efc3f8d3c38ffb3d945af9cd2531b96436e016dadR22-R38) [[2]](diffhunk://#diff-f632e2877b1f332884af3c5efc3f8d3c38ffb3d945af9cd2531b96436e016dadR244-R329) [[3]](diffhunk://#diff-f632e2877b1f332884af3c5efc3f8d3c38ffb3d945af9cd2531b96436e016dadR652-R699) [[4]](diffhunk://#diff-f632e2877b1f332884af3c5efc3f8d3c38ffb3d945af9cd2531b96436e016dadR721-R767)

**Template Model Fix**
* Fixed the template model to treat the `data` field as a string rather than a list of integers, aligning with the expected runner install script format.
#### Why we need it

Needed for tmate SSH server support.

#### Checklist

- [ ] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md`


charms/tests/integration/test_garm.py
```diff
@@ -501,3 +504,201 @@ def test_garm_api_has_configurator_provider(
...
+        "forge_type": "github",
+        "os_type": "linux",
+        "data": list(script.encode("utf-8")),
+    }
+    resp = requests.post(f"{base_url}/templates", json=payload, headers=headers, timeout=30)
+    resp.raise_for_status()
+    logger.info("Base template '%s' created", GARM_BASE_TEMPLATE_NAME)
+
+
+@pytest.fixture(scope="module", name="garm_with_debug_ssh")
```
florentianayuwono: should the fixtures be in conftest file?

florentianayuwono (commented):
thanks andrew! i haven't reviewed all the unit tests yet. i think there r q some overlaps w the garm sync scaleset for the integration tests fixture (rn im mocking it in my pr, so after the pr is merged u can perhaps replace those mocks w these real ones)

charms/garm/src/garm_api.py
```diff
@@ -140,3 +148,237 @@ def first_run(
...
+            Configured ApiClient with Bearer auth header.
+        """
+        config = Configuration(
+            host=self._base_url,
+            api_key={"Bearer": token},
+            api_key_prefix={"Bearer": "Bearer"},
+        )
+        return ApiClient(configuration=config)
+
+    def list_templates(self, token: str) -> list[Template]:
```
florentianayuwono: the authenticated api callers can be moved to be under the authenticated client class (implemented in my pr)

charms/tests/integration/test_garm.py
yhaliaw: This is where the other integration test fixtures is as well.

charms/garm/src/garm_api.py
yhaliaw: I am not seeing a new class: https://github.com/canonical/github-runner-operators/blob/main/charms/garm/src/garm_api.py

charms/garm/src/garm_client/models/template.py
```diff
@@ -29,7 +29,7 @@ class Template(BaseModel):
     Template
     """ # noqa: E501
     created_at: Optional[datetime] = None
-    data: Optional[List[StrictInt]] = None
+    data: Optional[StrictStr] = None
```
yhaliaw, line 32: Background: This is generated code and should normally not be modified.

Workaround for this issue: https://github.com/cloudbase/garm/issues/796

Will need to decide on how to proceed with this.

cbartz: Should we note this down somewhere, to avoid overrriding by re-generating?

charms/webhook-gateway-operator/lib/charms/squid_forward_proxy/v0/http_proxy.py
cbartz, line 1: this looks unrelated to this PR, is it required?

charms/garm/src/garm_api.py
```diff
@@ -579,3 +718,50 @@ def delete_scaleset(self, scaleset_id: int) -> None:
...
+        connections: List of SSHDebugInfo from the debug-ssh relation.
+
+    Returns:
+        A shell snippet string (no shebang) to be prepended to the base template,
+        or an empty string when there are no connections.
+    """
+    if not connections:
+        return ""
+    conn = connections[0]
+    runner_env = "/home/ubuntu/actions-runner/.env"
```
cbartz: should it rather be `/home/runner` ?

charms/garm/src/garm_api.py
yhaliaw: Updated.

charms/garm/src/garm_api.py
```diff
@@ -579,3 +718,50 @@ def delete_scaleset(self, scaleset_id: int) -> None:
                 ) from exc
             except urllib3.exceptions.HTTPError as exc:
                 raise GarmConnectionError(f"GARM connection error: {exc}") from exc
+
+
+def build_tmate_env_snippet(connections: list[SSHDebugInfo]) -> str:
```
cbartz: I think this fct lives in the wrong module, it should rather be in garm_template ?

cbartz (commented):
Thanks. Looks good overall. Some comments.

charms/garm/src/charm_state.py
```diff
@@ -0,0 +1,117 @@
...
+
+    connections: list[SSHDebugInfo] = []
+    for unit in relation.units:
+        data = relation.data[unit]
+        host = data.get("host")
+        port_str = data.get("port")
+        rsa_fingerprint = data.get("rsa_fingerprint")
+        ed25519_fingerprint = data.get("ed25519_fingerprint")
+
+        if not host or not port_str or not rsa_fingerprint or not ed25519_fingerprint:
```
florentianayuwono:
```suggestion
        if not (host and port_str and rsa_fingerprint and ed25519_fingerprint):
```

charms/garm/src/garm_template.py
```diff
@@ -0,0 +1,150 @@
...
+    return prepend_after_shebang(base_script, snippet).encode("utf-8")
+
+
+def _sync_charmed_template(
+    client: GarmAuthenticatedClient,
+    charmed: typing.Any,
+    patched_data: bytes,
+    n_connections: int,
+) -> int:
+    """Create or atomically update the charmed template; return its id."""
```
florentianayuwono, line 148: i think this function could benefit from having 2 helpers; 1 for updating and 1 for creating

charms/garm/src/scaleset_reconciler.py
```diff
@@ -217,4 +220,5 @@ def _needs_update(observed, spec: ScalesetSpec) -> bool:
             or observed.min_idle_runners != spec.min_idle_runners
             or observed.github_runner_group != (spec.runner_group or None)
             or observed_scripts != spec.pre_install_scripts
+            or (spec.template_id is not None and observed.template_id != spec.template_id)
```
florentianayuwono, line 223:
```suggestion
            or (spec.template_id and observed.template_id != spec.template_id)
```

charms/garm/tests/unit/test_garm_api.py
```diff
@@ -131,6 +137,80 @@ def test_login_returns_token():
...
+
+    @pytest.mark.parametrize(
+        "side_effect, expected, raises_match",
+        [
+            pytest.param(None, True, None, id="200-returns-true"),
+            pytest.param(ApiException(status=409), False, None, id="409-returns-false"),
+            pytest.param(ApiException(status=500), None, "500", id="500-raises"),
+        ],
+    )
+    def test_is_initialized(self, client, side_effect, expected, raises_match):
```
florentianayuwono: there's alr test_is_initialized before this?

charms/garm/tests/unit/test_garm_api.py
```diff
@@ -131,6 +137,80 @@ def test_login_returns_token():
...
+                mock_cls.return_value.controller_info.return_value = MagicMock()
+
+            if raises_match is not None:
+                with pytest.raises(GarmApiError, match=raises_match):
+                    client.is_initialized()
+            else:
+                assert client.is_initialized() is expected
+
+
+class TestFirstRun:
```
florentianayuwono: there's alr test first run and login too, perhaps we can refactor the tests in this file to avoid duplication

charms/garm/tests/unit/test_garm_template.py
```diff
@@ -0,0 +1,298 @@
...
+
+import garm_template
+from charm_state import SSHDebugInfo
+from garm_api import GarmApiError
+
+_CONN = SSHDebugInfo(host="h", port=10022, rsa_fingerprint="rsa-fp", ed25519_fingerprint="ed-fp")
+
+
+# ---------------------------------------------------------------------------
+# apply_charmed_template
```
florentianayuwono: perhaps we can remove these separators

charms/tests/integration/test_garm.py
florentianayuwono: i think the other fixtures r in conftest file? i didn't see any existing fixture here

florentianayuwono (commented):
thanks andrew!

charms/garm/src/garm_client/models/template.py
yhaliaw: I have patched the generation script to modify the `data` field prior to using OpenAPI generator.
Upstream GARM has merged the PR to fix this issue.

This will be a temp solution until will update to the latest GARM.

charms/garm/src/scaleset_reconciler.py
yhaliaw: For None checks it is generally good to be explicit.

Using truthiness checks might cause issues if the template_id is 0. I do not think GARM will ever use 0 as the template_id, but explicit None check is generally a good practice.

charms/tests/integration/test_garm.py
yhaliaw: I misread the original comment. I will move this to conftest.

charms/garm/tests/unit/test_garm_api.py
yhaliaw: You are right. Removed this test. Likely some merge conflict resolving done by AI duplicated the tests.

charms/garm/tests/unit/test_garm_api.py
yhaliaw: Done.

charms/garm/src/charm_state.py
yhaliaw: Good suggestion.

charms/garm/src/garm_api.py
yhaliaw: Move it and `prepend_after_shebang` to garm_template.py.

charms/garm/src/garm_template.py
yhaliaw: Done.

charms/garm/src/charm.py
```diff
@@ -352,6 +354,18 @@ def __init__(self, *args: typing.Any) -> None:
...
+        self.framework.observe(
+            self.on[DEBUG_SSH_INTEGRATION_NAME].relation_joined,
+            self._reconcile,
+        )
+        self.framework.observe(
+            self.on[DEBUG_SSH_INTEGRATION_NAME].relation_changed,
+            self._reconcile,
+        )
+        self.framework.observe(
+            self.on[DEBUG_SSH_INTEGRATION_NAME].relation_broken,
```
cbartz, line 375: `relation_departed` should be observed here too, routed to the same `_reconcile` — I'd treat this as a blocker given our tmate provider runs **multiple units**.

The charm injects only `connections[0]` (lowest `(host, port)` after sort), so exactly one tmate unit is ever written into `github_linux_charmed`. Scale-down of that selected unit fires only `relation-departed`, which isn't observed — so the template keeps pointing at the departed server's host/port until the next `update-status` reconcile. Runners spawned in that window get a tmate config for a server that no longer exists, so debug sessions to them fail until reconcile heals (default up to ~5 min, longer if `update-status-hook-interval` is raised).

Scale-*up* is fine (joined/changed are observed) and departure of a non-selected unit is harmless; the functional gap is specifically the selected unit departing. A one-line `relation_departed` observer closes it.

Separate, lower-priority design question this raises: with a multi-unit provider, is selecting only `connections[0]` intended, or should the template carry a failover set? Fine to defer, but worth a conscious call.

*This is an AI-generated message.*

charms/garm/scripts/generate_client.sh
```diff
@@ -21,12 +21,62 @@ PACKAGE_DIR="${OUTPUT_DIR}/garm_client"
...
 curl -fsSL "${SWAGGER_URL}" -o /tmp/garm_swagger.yaml
 
+# Patch the spec before generation. GARM renders the template `data` body (a Go []byte)
+# as `type: array, items: {type: integer, format: uint8}`, but the actual JSON wire format
+# is a base64 string. Rewrite the `data` property of the three template models to
+# `type: string` so the generated client uses StrictStr instead of List[int].
+# See https://github.com/cloudbase/garm/issues/796.
+# Scoped to template models on purpose: other []byte fields (e.g. private_key_bytes) are
+# still consumed as int-arrays elsewhere in the charm and must not be touched here.
+echo 'Patching template data (uint8-array) properties to strings...'
```
cbartz, line 31: The patch is correct and necessary at the currently pinned commit (`47811d0`, 2026-06-08) — the template `data` fields are still `uint8` arrays there. Just noting for a **follow-up**: upstream #802 (merged `fd712c6`, 2026-06-30) annotates `[]byte` as base64 strings in the spec, so bumping `GARM_COMMIT` past that and regenerating would let this whole patch block (and the `.patched.yaml` mount) be dropped — a nice simplification. `garm_api.py`'s base64 encode/decode stays regardless.

Two things to check before bumping so it's a conscious move, not a blind one: (1) the bump pulls ~3 weeks of unrelated API changes, so it's a full client regen + re-test, and (2) #802 annotates *most* byte arrays — confirm it didn't also flip a field the charm currently consumes as `List[int]` (e.g. `private_key_bytes`), which this patch deliberately leaves alone. Suggest keeping the patch in this PR and doing the bump/regen separately.

*This is an AI-generated message.*

charms/garm/src/charm.py
yhaliaw: I have added the relation_departed event.

charms/garm/src/garm_template.py
```diff
@@ -0,0 +1,208 @@
...
+    Returns:
+        A shell snippet string (no shebang) to be prepended to the base template,
+        or an empty string when there are no connections.
+    """
+    if not connections:
+        return ""
+    conn = connections[0]
+    runner_env = "/home/runner/.env"
+    lines = [
+        f"mkdir -p $(dirname {runner_env})",
```
cbartz, line 78: Question, not a change request: this writes `/home/runner/.env` as root at the top of the install script (before the `runner` user/home exist), so the file ends up root-owned. The values here are public (host/port/fingerprints), so root-owned + world-readable is harmless **if** the file is only ever read. Does anything append to `/home/runner/.env` later as the runner user? If not, this is fine as-is. If it does, it's worth writing to a root-owned location the runner already sources (systemd `EnvironmentFile` / drop-in) or chowning after the user is created.

_AI-generated message._

charms/garm/src/charm.py
```diff
@@ -884,8 +915,13 @@ def _reconcile_runners(self) -> None:
             self._ensure_controller_urls(auth_client)
             GithubReconciler(auth_client).reconcile(self._build_desired_credentials())
             EntityReconciler(auth_client).reconcile(CharmState.from_charm(self).desired_entities)
-            ScalesetReconciler(auth_client).reconcile(self._build_desired_scalesets())
+            connections = CharmState.from_charm(self).ssh_debug_connections
```
cbartz: Minor: `CharmState.from_charm(self)` is built here and again on the previous line for `desired_entities`, re-reading all relations twice. Could build once and reuse:
```python
state = CharmState.from_charm(self)
EntityReconciler(auth_client).reconcile(state.desired_entities)
template_id = _apply_garm_template(auth_client, state.ssh_debug_connections)
```

_AI-generated message._

charms/garm/AGENTS.md
```diff
@@ -9,11 +9,18 @@ charm conventions; this file lists only what's specific to `garm`.
 - **Secrets (owner).** `garm` owns two labelled juju secrets — `GARM_SECRETS_LABEL` and `GARM_ADMIN_CREDENTIALS_LABEL` — created leader-only in `_ensure_secrets`.
   - **DO** read them with plain `get_content()` (`_get_secrets`, `_get_admin_credentials`).
   - **DON'T** pass `refresh=True` — that's an observer concept (see root `AGENTS.md`).
+- **Domain logic is factored out of `charm.py`**: `src/garm_api.py` and `src/garm_client/`
```
cbartz, line 12: Nit: this bullet is duplicated — the identical line follows on line 13. Looks accidental; drop one.

_AI-generated message._

charms/garm/src/garm_api.py
```diff
@@ -759,6 +850,54 @@ def update_scaleset(self, scaleset_id: int, params: UpdateScaleSetParams) -> Sca
...
+                    _request_timeout=_REQUEST_TIMEOUT,
+                )
+            except ApiException as exc:
+                raise GarmApiError(
+                    f"GARM update_template({template_id}) failed ({exc.status}): {exc.body}"
+                ) from exc
+            except urllib3.exceptions.HTTPError as exc:
+                raise GarmConnectionError(f"GARM connection error: {exc}") from exc
+
+    def delete_template(self, template_id: int) -> None:
```
cbartz, line 881: Nit: `delete_template` doesn't appear to be called anywhere — the charmed template is only ever created/updated. Either wire it into a cleanup path or drop it to avoid untested-by-need API surface. If it's kept intentionally for symmetry, a one-line note on intent would help.

_AI-generated message._

cbartz (approved):
Nice work — clean design and well tested. Parsing debug-ssh data into immutable `CharmState`, deriving `github_linux_charmed` from GARM's built-in base, base64-compare before update, the newline-injection guard, and stable sorting are all handled thoughtfully. Unit suite passes locally (182 tests).

No blockers. One question and a few nits below, none of which need to gate the merge.

One thing worth calling out in the description for future readers: every deployment now gets a `github_linux_charmed` template (a snippet-free copy of base when no debug-ssh relation) and all scalesets reference it, plus a few unconditional template API calls on every reconcile including the 5-min `update_status`. Looks intended per the design — just making it explicit.

_This review was generated with AI assistance._

charms/garm/AGENTS.md
yhaliaw: This seems to not be introduced in this PR. I will do a cleanup PR after this.

charms/garm/src/garm_template.py
yhaliaw: The .env file in this directory is read by GitHub runner application. I think the application does not pick up env var from other systemd Environment file.
And the file is only meant for reading.

charms/garm/src/charm.py
yhaliaw: Updated.

charms/garm/src/garm_api.py
yhaliaw: I removed it, since it is unused code.
Weird that my code-review agent did not pick this up as the prompt includes unused code detection.

PR #254, opened 2026-06-24
feat(garm): add get-credentials action

Thanhphan1147:
## Summary

- Adds the `get-credentials` Juju action to the GARM charm, exposing the admin credentials stored in the `garm-admin-credentials` Juju secret to the operator
- Adds unit tests covering the success path (credentials returned via \`event.set_results\`) and the failure path (credentials not yet available, \`event.fail\` called)
- Adds an integration test that runs the action against a live GARM unit and asserts all four credential fields are returned
- Adds a how-to guide under \`docs/how-to/retrieve-garm-credentials.md\`

## Usage

```
juju run garm/0 get-credentials
```


cbartz (approved):
LGMT. This was not planned, but seems indeed like a useful feature.

charms/tests/integration/test_garm.py
```diff
@@ -501,3 +501,33 @@ def test_garm_api_has_configurator_provider(
...
+    assert: The action succeeds and returns all four credential fields (username,
+        password, email, full-name), with username "admin" and email "admin@garm.local".
+    """
+    unit = f"{garm_app}/0"
+    logger.info("Running get-credentials action on unit %s", unit)
+
+    # juju.run() raises TaskError if the action fails, so a clean return means success.
+    task = juju.run(unit, "get-credentials")
+
+    logger.info("get-credentials results: %s", dict(task.results))
```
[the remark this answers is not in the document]
Thanhphan1147: @copilot remove this logger line entirely

charms/garm/tests/unit/test_charm.py
```diff
@@ -540,3 +540,43 @@ def test_maybe_first_run_skips_on_missing_credential_key():
...
+        "password": "s3cr3t",
+        "email": "admin@garm.local",
+        "full-name": "GARM Admin",
+    }
+    charm._get_admin_credentials.return_value = credentials
+
+    GarmCharm._on_get_credentials_action(charm, event)
+
+    event.set_results.assert_called_once_with(credentials)
+    event.fail.assert_not_called()
```
florentianayuwono, line 548: can we add assertion that the credentials returned is as expected

.gitignore
```diff
@@ -43,4 +43,7 @@ build/
 
 # Editor/IDE
 .idea/
-.vscode/
\ No newline at end of file
+.vscode/
+
+# Git worktrees
+.worktrees/
```
florentianayuwono: add eof here?

florentianayuwono (commented):
thank you!

.gitignore
Thanhphan1147: @copilot Add a trailing new line

charms/garm/tests/unit/test_charm.py
Thanhphan1147: It's what the `event.set_results.assert_called_once_with(credentials)` is doing. Since we're mocking the event and not using scenario we don't have a way to actually get the output ( the hook methods always return None )

florentianayuwono: i see, is there a reason why we cannot use scenario? ideally it's better to test the state, not interaction

Thanhphan1147: No particular reason aside from that I like to keep tests simple if it's only a single method. And using scenario we'd still only call the hook method so it's a bit more overhead. But I can switch to scenario if needed, wdyt?

charms/tests/integration/test_garm.py
```diff
@@ -520,6 +520,35 @@ def garm_login_from_secret(juju: jubilant.Juju, garm_app_name: str, garm_url: st
...
+        assert key in task.results, (
+            f"Expected '{key}' in action results, got: {list(task.results)}"
+        )
+    assert task.results["username"] == "admin", (
+        f"Expected username 'admin', got: {task.results['username']!r}"
+    )
+    assert task.results["email"] == "admin@garm.local", (
+        f"Expected email 'admin@garm.local', got: {task.results['email']!r}"
+    )
+    assert task.results["password"], "Expected non-empty password in action results"
```
Thanhphan1147: @copilot  Move this section to below the file as it's inserting itself in the middle of another test

charms/garm/charmcraft.yaml
```diff
@@ -32,3 +32,9 @@ requires:
     limit: 1
   garm-configurator:
     interface: garm_configurator_v0
+    limit: 1
```
Thanhphan1147, line 37: @copilot remove this as it's not part of the PR

docs/how-to/retrieve-garm-credentials.md
```diff
@@ -0,0 +1,37 @@
+# How to retrieve GARM admin credentials
+
+The `get-credentials` action retrieves the GARM admin credentials generated by the charm during start-up. Use it when you need to authenticate via the GARM API or when registering new profiles using `garm-cli`.
+
+## Prerequisites
+
+- The GARM charm must be fully initialised. The unit status must be `active` before running this action. If the charm has not completed its first-run setup, the credentials will not yet be available and the action will fail.
```
erinecon:
```suggestion
The GARM charm must be fully initialized. The unit status must be `active` before running this action. If the charm has not completed its first-run setup, the credentials will not yet be available and the action will fail.
```
Nits, for US spelling and also to remove the bulleted formatting since there's only one prerequisite

docs/how-to/retrieve-garm-credentials.md
```diff
@@ -0,0 +1,37 @@
...
+| `password` | The GARM admin password. |
+| `email` | The email address associated with the admin account. |
+| `full-name` | The full name associated with the admin account. |
+
+Example output:
+
+```
+Running operation 1 with 1 task
+  - task 2 on unit-garm-0
+
```
erinecon:
````suggestion
```{terminal}
:output-only:

Running operation 1 with 1 task
  - task 2 on unit-garm-0

````
nit, also be careful with using GitHub's auto-incorporating feature since there are three backticks in the suggestion

**Review by erinecon — APPROVED** [review]
Thanks so much for adding a how-to guide!! Some nits from me. Giving an approval since @cbartz has approved already

**Review by yhaliaw — APPROVED** [review]

File: charms/garm/tests/unit/test_charm.py [hunk 8]
```diff
@@ -520,6 +520,44 @@ def test_maybe_first_run_skips_on_missing_credential_key():
     mock_client_cls.assert_not_called()
 
 
+def test_get_credentials_action_returns_credentials_when_available():
+    """
+    arrange: Admin credentials secret exists and contains valid content.
+    act: Call _on_get_credentials_action().
+    """
```
yhaliaw, line 534: This docstring is missing its `assert:` section — it only has `arrange:`/`act:`, while the test body performs two assertions. The sibling test added in the same commit (`test_get_credentials_action_fails_when_credentials_unavailable`) includes all three sections, as does every other test in this file.

Suggestion:
```suggestion
    assert: event.set_results is called once with the credentials dict and
        event.fail is not called.
    """
```

charms/garm/tests/unit/test_charm.py
```diff
@@ -520,6 +520,44 @@ def test_maybe_first_run_skips_on_missing_credential_key():
...
+    charm = MagicMock()
+    event = MagicMock()
+    charm._get_admin_credentials.return_value = None
+
+    GarmCharm._on_get_credentials_action(charm, event)
+
+    event.fail.assert_called_once()
+    fail_message = event.fail.call_args[0][0]
+    assert "not yet available" in fail_message
+    event.set_results.assert_not_called()
```
yhaliaw, line 567: There are zero blank lines between the end of this test and the next top-level `def test_reconcile_scalesets_skips_when_no_admin_credentials():` — PEP 8 / ruff `E302` expects two. This won't fail CI (this charm's `tox -e lint` only checks `src`, not `tests/`), but it will get reformatted the next time someone runs `tox -e fmt`, producing unrelated churn in a future PR.

```suggestion
    event.set_results.assert_not_called()


```

yhaliaw (commented):
Posting three minor findings from a code review pass — all cosmetic/convention issues, none blocking.

charms/garm/tests/unit/test_charm.py
florentianayuwono: i think for now it's okay to do it like this. the team will migrate the entire tests to scenario in the future

PR #255, opened 2026-06-26
feat(garm): make all hooks flow through a single reconcile

cbartz:
#### What this PR does

Replaces the four separate garm charm hook handlers (_on_install, _on_leader_elected, _on_update_status, _on_configurator_relation_changed) with a single _reconcile dispatcher that calls restart(). Moves _ensure_secrets() above the is_ready() gate in restart() so secrets are created on install/leader_elected before pebble is ready. No longer overrides _on_update_status — the base class observer handles ingress refresh and migration-failed restart; the garm _reconcile observer runs alongside it.

#### Why we need it

The garm charm was not fully holistic: install and leader_elected only called _ensure_secrets() without reconciling, and update_status called _reconcile_scalesets() directly, bypassing restart(). Now every hook flows through the same restart() entry point, and the TOML hash-skip inside restart() prevents unnecessary Pebble restarts on periodic update_status ticks.

Updated AGENTS.md (root, garm, planner-operator) to clarify that restart() is the reconcile and a thin _reconcile dispatcher is acceptable.

#### Checklist

- [x] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant AGENTS.md

#### Test plan

- Unit tests: tox -e unit (77 passed), tox -e lint, tox -e complexity, tox -e static — all green.
- Integration tests: no changes needed — black-box tests assert on statuses, secrets, API, and scalesets, all preserved.

#### Review focus

- The _reconcile dispatcher pattern vs the AGENTS.md guidance about not adding _reconcile. The guidance is about not implementing competing reconcile logic; a thin dispatcher that just calls restart() is DRY, not a competing reconcile.
- _ensure_secrets() moved above is_ready() gate — it only needs is_leader() + Juju secret store, no container access.
- update_status no longer calls super()._on_update_status() — the base class registers its own observer in __init__, so both fire independently.

### DIFF
--- a/charms/garm/src/charm.py
+++ b/charms/garm/src/charm.py
@@ -274,48 +274,30 @@ class GarmCharm(paas_charm.go.Charm):
             args: Passed through to CharmBase.
         """
         super().__init__(*args)
-        self.framework.observe(self.on.install, self._on_install)
-        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(self.on.install, self._reconcile)
+        self.framework.observe(self.on.leader_elected, self._reconcile)
         self.framework.observe(
             self.on[GARM_CONFIGURATOR_RELATION_NAME].relation_joined,
-            self._on_configurator_relation_changed,
+            self._reconcile,
         )
         self.framework.observe(
             self.on[GARM_CONFIGURATOR_RELATION_NAME].relation_changed,
-            self._on_configurator_relation_changed,
+            self._reconcile,
         )
         self.framework.observe(
             self.on[GARM_CONFIGURATOR_RELATION_NAME].relation_departed,
-            self._on_configurator_relation_changed,
+            self._reconcile,
         )
         self.framework.observe(
             self.on[GARM_CONFIGURATOR_RELATION_NAME].relation_broken,
-            self._on_configurator_relation_changed,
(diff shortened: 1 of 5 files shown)


yanksyoon (approved):

charms/garm/src/charm.py
```diff
@@ -274,48 +274,30 @@ def __init__(self, *args: typing.Any) -> None:
...
-        on the periodic poll. Calls _reconcile_scalesets() directly to
-        avoid triggering an unnecessary Pebble service restart — TOML
-        changes are already handled by the events that caused them.
-
-        Args:
-            event: Juju update-status event.
-        """
-        super()._on_update_status(event)
-        self._reconcile_scalesets()
+    def _reconcile(self, _: ops.EventBase) -> None:
```
javierdelapuente, line 298: Why _reconcile that calls restart and not calling restart?

if at some point code is added before or after the restart in this function, the code will not be holistic.

The name restart is historical, and may not be the best one, but it is the entry point to the reconcile called in lots of handlers, using another entry point here introduce risks.

cbartz: @javierdelapuente Because restart is overriding the parent and has a different signature. Also the @block_if_config_invalid decorator would not be applicable to restart.

PR #265, opened 2026-07-06
feat(garm): support runner config via scaleset templates

cbartz:
#### What this PR does

- adds runner-level configuration plumbing from `garm-configurator` into GARM scaleset reconciliation
- renders per-scaleset runner-install templates for Docker mirror, proxy/aproxy, OTEL, and pre-job hooks
- adds unit and integration coverage for the new runner-template flow, plus the test doc cleanup in the new coverage

#### Why we need it

GARM stores runner-install behavior in templates, so exposing runner options from the configurator is only useful if the operator can reconcile those options into the scaleset's effective template without a manual upgrade or restart cycle.

#### Checklist

- [x] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md`

#### Test plan

- `tox -c charms/garm/tox.toml -e unit` (`199 passed`)
- `pytest charms/garm/tests/unit/test_runner_template.py charms/garm/tests/unit/test_scaleset_reconciler.py` (`30 passed`)
- `tox -c charms/garm-configurator/tox.toml -e unit` currently fails during tox env setup because `ops-scenario==8.7.1` depends on `ops==3.7.1` while `requirements.txt` pins `ops==3.8.0`

#### Review focus

- `RunnerConfig` validation and relation plumbing in `garm-configurator`
- template lifecycle handling in `scaleset_reconciler.py`, especially create/update/detach behavior
- rendered runner-install script contents in `runner_template.py`

- no API break intended; this changes how runner options are materialized by creating per-scaleset GARM templates

#### New dependencies, APIs, modules, or workflow changes

- no new external runtime dependencies
- adds the new `charms/garm/src/runner_template.py` module
- extends the configurator-to-garm relation contract with runner-level config fields
- adds integration coverage in `charms/tests/integration/test_garm.py`


charms/garm/src/runner_template.py
```diff
@@ -0,0 +1,515 @@
...
+"""
+
+import ipaddress
+import json
+import re
+import shlex
+
+from charm_state import RunnerConfig
+
+_PORT_TOKEN_RE = re.compile(r"^\d{1,5}(-\d{1,5})?$")
```
yanksyoon: Would this be still necessary to validate things here? I think we can treat everything that was parsed from charm state as validated. WDYT?

charms/garm/src/runner_template.py
```diff
@@ -0,0 +1,515 @@
...
+RUNNER_HOME = "/home/runner/actions-runner"
+PRE_JOB_HOOK_PATH = f"{RUNNER_HOME}/pre-job.sh"
+RUNNER_ENV_PATH = f"{RUNNER_HOME}/env"
+
+# OpenTelemetry collector config: hostmetrics/otlp/loki pipelines exporting to
+# the configured endpoint. `$GITHUB_*`, `$RUNNER_NAME` and `$(uname -m)` are left
+# as literal shell syntax — they sit inside the *inner*, unquoted `tee <<EOF`
+# heredoc, so they only expand when the runner executes this hook, once per job.
+# Only the exporter endpoint is substituted at render time.
+_OTEL_COLLECTOR_SETUP = """\
```
yanksyoon: I'm wondering if this should be a J2 template rather than hardcoded string template, but since it's a rock, the effort of updating the templates in the rock might not be worth it.

yanksyoon: I'm wondering if this should be a J2 template rather than hardcoded string template, but since it's a rock, the effort of updating the templates in the rock might not be worth it.

charms/garm/src/runner_template.py
```diff
@@ -0,0 +1,515 @@
...
+        base: The preferred delimiter; extended only if it collides.
+
+    Returns:
+        ``base``, suffixed with underscores until no line of *content* matches it.
+    """
+    lines = content.splitlines()
+    delimiter = base
+    while delimiter in lines:
+        delimiter += "_"
+    return delimiter
```
yanksyoon: I feel that such processing helpers indicate that it miiight be better as a j2 template

yanksyoon (approved):
LGTM, a discussion point about templating, thanks @cbartz !

charms/garm/src/runner_template.py
cbartz: Good call — moved the port/address/env-value sanitisation into `RunnerConfig.from_databag` (commit cd75155), so a parsed `RunnerConfig` is already clean and the renderer just consumes it. I kept the re-validation on the garm side rather than trusting the configurator's validation alone, since the databag is written by a separate charm; the renderer now only does shell-safe embedding (quoting, heredoc-delimiter selection).

_AI-generated message._

charms/garm/src/runner_template.py
cbartz: Done — the renderer now builds from Jinja2 templates in `src/templates/*.j2` (loaded via `FileSystemLoader`) instead of string concatenation (commit cd75155).

On the rock concern: I verified `src/` is packed verbatim into the charm (the built charm already ships nested non-`.py` files), so the `.j2` files travel with the charm code — no separate rock-update step. Rendered output is byte-for-byte identical (pinned by the unit tests).

_AI-generated message._

charms/garm/src/runner_template.py
cbartz: These are down to just the shell-safety helpers now (heredoc-delimiter selection, `shlex.quote`); the value validation moved to `RunnerConfig.from_databag` and the block bodies themselves moved to `src/templates/*.j2` as above.

_AI-generated message._

PR #272, opened 2026-07-08
fix(garm): deliver aproxy bootstrap as pre-install script so it runs before GARM's install wrapper

cbartz:
#### What this PR does

Delivers the aproxy proxy bootstrap as a GARM `pre_install_scripts` extra-spec entry (`00-aproxy`) instead of injecting it into the per-scaleset runner template. When a proxy is configured, cloud-init package upgrades are also disabled on boot. Two related fixes: `snap set aproxy proxy=` now gets a bare `host:port` (aproxy rejects URLs), and all `pre_install_scripts` values are base64-encoded as GARM's `map[string][]byte` extra-spec type requires.

#### Why we need it

On proxy-only networks (no direct egress), runners never registered. When a scaleset uses a server-side template, GARM prepends a compiled-in wrapper as the cloud-init install script; that wrapper must reach the GARM metadata API to fetch the template *before* any template content runs. Proxy bootstrap living inside the template therefore can never help the wrapper's fetch succeed — the fetch times out (~134 s/attempt) with no egress. GARM's `pre_install_scripts` run as root *before* the wrapper, so moving aproxy there brings egress up in time. The apt upgrade runs even earlier (before any pre-install script), so it's disabled when a proxy is set rather than left to time out.

#### Test plan

- Unit: retargeted the aproxy rendering tests at the new `render_aproxy_pre_install_script`, added happy-path + scheme-stripping (`http://` / `https://` / scheme-less → `host:port`) coverage, and added scaleset-reconciler tests for the `00-aproxy` + `disable_updates` create path and three drift cases (encoded match → no-op; legacy raw-text → update; missing `disable_updates` → update). `tox -c tox.toml -e fmt,lint,static,unit` all pass; touched-module coverage 98% / 89%.
- Integration: updated `test_runner_options_render_into_scaleset_template` to assert the aproxy script now lands in the scaleset's `extra_specs` (base64-decoded `00-aproxy`, `disable_updates: true`) rather than the template. Not run here (needs a live Juju model).

#### Review focus

- `_effective_extra_specs()` in `scaleset_reconciler.py` is the single source of truth feeding create, update, and drift detection — worth confirming the drift comparison (encoded scripts + `disable_updates`) is symmetric with what's sent.
- The `00-aproxy` key name is load-bearing: GARM runs pre-install scripts in lexicographic order, so it must sort before the configurator's `pre_install.sh`.

Existing scalesets get a one-time update on first reconcile as their `extra_specs` converge (encoding change + aproxy entry). No config-surface change. Operational prerequisite: the egress proxy must allow CONNECT to the GARM API host, since all GARM API traffic from runners now rides through aproxy.

#### Checklist

- [ ] `CONTRIBUTING.md` has been updated upon changes to the contribution/development process (e.g. changes to the way tests are run) — n/a
- [ ] Technical author has been assigned to review the PR in case of documentation changes (usually *.md files) — n/a (changelog only)
- [ ] **If integration test modules are used:** I updated the workflow configuration — n/a (existing `test_garm` module, no new module)
- [ ] **If this PR involves a Grafana dashboard:** I added a screenshot of the dashboard — n/a
- [ ] **If this PR involves Terraform:** `terraform fmt` passes and `tflint` reports no errors — n/a
- [ ] **If this PR involves Rockcraft:** I updated the version — n/a
- [ ] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md` — n/a (no convention change)
- [ ] **If this PR changes `.copilot-collections.yaml` or `.github/instructions/`:** I re-checked the `AGENTS.md` "12-factor divergences" guidance — n/a


charms/garm/src/scaleset_reconciler.py
```diff
@@ -431,14 +429,52 @@ def _maybe_update(self, observed: ScaleSet, spec: ScalesetSpec, template_id: int
...
+        ``pre_install_scripts`` as ``map[string][]byte``); empty when the spec
+        yields no extra specs.
+    """
+    scripts = dict(spec.pre_install_scripts)
+    extra_specs: dict[str, object] = {}
+    if spec.runner_config.runner_http_proxy:
+        # The aproxy bootstrap must run before GARM's compiled-in install wrapper
+        # (which needs egress to fetch the runner template), hence a pre-install
+        # script rather than template content.
+        scripts[APROXY_SCRIPT_NAME] = render_aproxy_pre_install_script(spec.runner_config)
```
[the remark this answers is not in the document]
cbartz: Declining: operator-supplied pre-install scripts reach the reconciler only via garm-configurator, which always writes them under the single hardcoded key `pre_install.sh` (`json.dumps({"pre_install.sh": ...})`). The `00-aproxy` namespace therefore can't collide with an operator key in practice, and adding a fail-fast would risk turning a structurally-impossible case into a blocked-status path. The `00-` numeric prefix already documents the slot as charm-reserved.

_AI-generated message on behalf of the PR author._

charms/garm/src/templates/aproxy.j2
```diff
@@ -21,4 +48,12 @@ table ip aproxy {
...
\ No newline at end of file
+
+# Only redirect once aproxy is actually listening; otherwise the DNAT would
+# black-hole all :80/:443 egress by sending it to a dead port.
+if snap services aproxy 2>/dev/null | grep -q active; then
+    log "aproxy is listening; applying nftables redirect"
+    nft -f /etc/nftables.conf
+else
+    log "ERROR: aproxy service is not active; leaving egress unchanged (no redirect)"
+fi
```
yanksyoon: out of curiosity, would this require timeout/retries? from the description it sounds like this is a process in which we "wait" to become active.

charms/garm/src/scaleset_reconciler.py
```diff
@@ -431,14 +436,53 @@ def _maybe_update(self, observed: ScaleSet, spec: ScalesetSpec, template_id: int
...
             or observed.flavor != spec.flavor
             or observed.max_runners != spec.max_runners
             or observed.min_idle_runners != spec.min_idle_runners
             or observed.enabled is not True
             or observed.github_runner_group != (spec.runner_group or None)
-            or observed_scripts != spec.pre_install_scripts
+            or observed_extra.get("pre_install_scripts", {})
+            != desired_extra.get("pre_install_scripts", {})
+            or bool(observed_extra.get("disable_updates"))
+            != bool(desired_extra.get("disable_updates"))
```
yanksyoon, line 453: Just curious about the bool wrapper for this field

charms/garm/tests/unit/test_scaleset_reconciler.py
```diff
@@ -359,15 +359,152 @@ def test_pre_install_scripts_passed_in_create():
...
+    assert: No update is issued.
+    """
+    from charm_state import RunnerConfig
+
+    proxy = "http://squid.internal:3128"
+    spec = _spec(
+        pre_install_scripts={"pre_install.sh": "echo operator-script"},
+        runner_config=RunnerConfig(runner_http_proxy=proxy),
+    )
+    from scaleset_reconciler import _effective_extra_specs
```
yanksyoon: Could imports be at the top of the file for organisation?

yanksyoon (approved):
Minor nits, thank you @cbartz !

charms/garm/src/templates/aproxy.j2
```diff
@@ -21,4 +48,12 @@ table ip aproxy {
...
 EOF
 systemctl enable nftables.service
-nft -f /etc/nftables.conf
\ No newline at end of file
+
+# Only redirect once aproxy is actually listening; otherwise the DNAT would
+# black-hole all :80/:443 egress by sending it to a dead port.
+if snap services aproxy 2>/dev/null | grep -q active; then
+    log "aproxy is listening; applying nftables redirect"
+    nft -f /etc/nftables.conf
```
weiiwang01: If this is a one time use machine that would never see a reboot, then writing the `nftables.conf` and applying it using the `nft -f /etc/nftables.conf` command is unnecessary, as you can directly pipe the script to `nft -f -`. Also, `systemctl enable nftables.service` is unnecessary for a one time use machine.

weiiwang01 (approved):

charms/garm/src/templates/aproxy.j2
cbartz: Good question — it reads like a poll, but it's really just a confirmation. `snap install` and `snap set` are synchronous: snapd blocks until the snap is installed and its service is started, so by the time control reaches this check the daemon is already up. There's also natural slack — nothing here consumes the redirect immediately; its consumer is the install wrapper's `curl`, which runs as a *later* cloud-init runcmd.

The one caveat is that `snap services … active` reflects systemd's view of the daemon, not that aproxy has bound `:54969` yet. If we ever observe aproxy being slow to bind in the field, a bounded poll on the port (`ss -ltn | grep :54969`) would be strictly more correct than this service-state snapshot — but given the boot-time slack I've left it as a single check for now.

> _AI-generated message._

charms/garm/src/scaleset_reconciler.py
cbartz: It normalises three cases into one comparison so drift detection doesn't flap: the flag being absent (`.get` → `None`) vs present-and-`True`, GARM round-tripping it as a truthy non-bool (e.g. a JSON string) rather than a real bool, and symmetry with the desired side, which only ever sets `True` or omits the key. Without the `bool()`, `None != True` (or `"true" != True`) would report "needs update" on every reconcile and loop forever — the same convergence trap as the `extra_specs` clear-to-empty case.

> _AI-generated message._

charms/garm/src/templates/aproxy.j2
cbartz: Good catch — you're right, these runner VMs are single-use and never reboot, so persisting the ruleset to a file and enabling `nftables.service` was dead weight. Fixed in c6a09e5: the ruleset is now piped straight into the kernel with `nft -f - << EOF`, and the `systemctl enable nftables.service` line is gone. Kept the `snap services aproxy … active` guard around it so we still never DNAT into a dead port.

> _AI-generated message._

charms/garm/tests/unit/test_scaleset_reconciler.py
cbartz: Agreed — hoisted them in c6a09e5. `RunnerConfig`, `_effective_extra_specs` and `build_template_data` are now imported at the top of the file, and all the function-local copies are removed.

> _AI-generated message._

PR #274, opened 2026-07-09
fix(garm): run injected pre-install steps with sudo

cbartz:
#### What this PR does

Prefixes the charm-injected pre-install steps (Docker registry mirror + host prep) with `sudo` so they actually run on the runner.

GARM runs the per-scaleset install script as the unprivileged `runner` user (`su -l -c ... runner`), and the base template escalates every privileged step with `sudo`. The two injected blocks omitted it, so on a real runner they failed silently:

- `dockerhub_mirror.j2` couldn't write `/etc/docker/daemon.json` or restart Docker → the registry mirror was never applied.
- `static_host_prep.j2` couldn't `adduser runner lxd/adm` → the runner never joined those groups.

The failures were invisible because the injected block runs *before* the base template's `set -e`, so the runner still registered. The `daemon.json` write uses `sudo tee` because a plain `sudo cat > file` still performs the redirect in the unprivileged shell. The misleading "runs as root" comment/docstrings that led to the omission are corrected.

#### Why we need it

Observed on latest `main`: a runner came up "successful" but its cloud-init log showed `/etc/docker/daemon.json: Permission denied`, `Interactive authentication required`, and `Only root may add a user`. The mirror and `lxd`/`adm` membership were silently dropped. The sibling `otel_collector_setup.j2` block already uses `sudo` correctly — this brings the two pre-install blocks in line.

#### Test plan

- Unit: `tox -c tox.toml -e unit` (221 passed). Assertions in `test_runner_template.py` were tightened first (TDD, RED→GREEN) to pin the `sudo` escalation.
- `tox -c tox.toml -e lint` and `-e static` (pyright) pass; `make -C docs spellcheck` clean.
- End-to-end (recommended before merge): repack + redeploy, spin a runner with a `dockerhub_mirror` set, confirm the cloud-init log is clean and `cat /etc/docker/daemon.json` / `groups runner` reflect the mirror and `lxd`/`adm`.

#### Review focus

- `sudo tee` vs `sudo cat >` for the `daemon.json` heredoc write — the redirect must be inside the privileged process.
- No change to `pre_job_hooks.j2` (writes runner-owned paths, correctly no `sudo`) or the aproxy path (already root via `pre_install_scripts`).
- No breaking changes, new dependencies, or API/workflow changes.

#### Checklist

- [ ] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md` 
- [ ] **If this PR changes `.copilot-collections.yaml` or `.github/instructions/`:** I re-checked the `AGENTS.md` "12-factor divergences" guidance 

🤖 Generated with [Claude Code](https://claude.com/claude-code)


charms/garm/src/templates/dockerhub_mirror.j2
```diff
@@ -1,7 +1,7 @@
...
-cat > /etc/docker/daemon.json <<'GARM_CHARM_DOCKER'
+sudo mkdir -p /etc/docker
+sudo tee /etc/docker/daemon.json > /dev/null <<'GARM_CHARM_DOCKER'
 {{ daemon_json }}
 GARM_CHARM_DOCKER
-systemctl daemon-reload
-systemctl restart docker
\ No newline at end of file
+sudo systemctl daemon-reload
+sudo systemctl restart docker
```
yanksyoon: Nit: missing newlines! (on other files as well)

yanksyoon (approved):

yhaliaw (approved):

charms/garm/src/templates/dockerhub_mirror.j2
cbartz: Good catch — fixed in d02614a. Added the trailing newline to both `dockerhub_mirror.j2` and `static_host_prep.j2` (the two fragment templates in this diff that were missing one).

These fragments deliberately had no final newline because `pre_install.j2` composes them and `keep_trailing_newline=True`, so a naive newline would inject a blank line between blocks. To keep the rendered install script byte-identical I compensated with Jinja whitespace control (`{{ dockerhub -}}` / `{{ static_prep -}}`) — verified the output is unchanged.

Note: a few other fragment templates that predate this PR (`hook_body.j2`, `otel_collector_setup.j2`, `custom_pre_job_script.j2`) also lack a final newline, but they're outside this PR's scope. Happy to clean those up in a follow-up if you'd like.

---
*This is an AI-generated message.*

cbartz: Follow-up in 56040a3: added the final newline to the remaining fragment templates too (`hook_body.j2`, `otel_collector_setup.j2`, `custom_pre_job_script.j2`), so all of them are consistent now.

To keep the rendered install script byte-identical, I unified the compensation into a single `_render_fragment()` helper that strips a fragment's trailing newline at splice time (replacing the per-template `-}}` whitespace-control). The composing template still owns the spacing between blocks. Verified byte-identical output across every runner-option combination.

---
*This is an AI-generated message.*

PR #281, opened 2026-07-12
fix(garm): reconcile scalesets when garm-configurator relation is removed

cbartz:
#### What this PR does

Deletes GARM scalesets when the `garm-configurator` relation is removed. `restart()` short-circuited on the now-empty provider configs and returned before `_reconcile_runners()` — the only path to `ScalesetReconciler`'s orphan-delete — ever ran, so scalesets whose desired spec left with the relation stayed registered in GARM (still shown by `garm-cli scaleset list`). The charm now runs the runner reconcile even when no configurator relation is present.

#### Why we need it

Removing an integration should tear down what it created. Without this, orphaned scalesets linger in GARM indefinitely after the relation is gone — consuming quota and continuing to appear in `garm-cli scaleset list` with no way for the charm to clean them up.

#### Test plan

- New unit test `test_restart_without_configurator_relation_still_prunes_orphaned_scalesets`: asserts `_reconcile_runners` runs, no Pebble replan happens (`add_layer` not called), and status stays `Waiting for garm-configurator relation`.
- Updated `test_restart_missing_configurator_relation_refreshes_stale_app_status` to stub `_reconcile_runners` (it builds the charm via `object.__new__` with no real `self.model`); its original status assertions are unchanged.
- `tox -c tox.toml -e unit` (223 passed), plus `fmt`, `lint`, `static` (pyright, 0 errors).

#### Review focus

- **Same drain-then-delete semantics, now reachable.** `ScalesetReconciler` already disables (`enabled=False`, `min_idle_runners=0`) then deletes, and catches GARM's 400 for a scaleset with active runners, retrying on the next reconcile (`update-status`). This PR doesn't change that logic — it just makes it run on the relation-removed path.
- **The fix is the one added `_reconcile_runners()` call** before the existing early return; the comment explains why the reconcile must still run with no relation present.

None to config or interfaces. Behaviour change only: scalesets orphaned by removing the `garm-configurator` relation are now cleaned up instead of stranded.

#### Checklist

- [ ] `CONTRIBUTING.md` has been updated upon changes to the contribution/development process (e.g. changes to the way tests are run) — N/A, no process change
- [ ] Technical author has been assigned to review the PR in case of documentation changes (usually *.md files) — only a routine `changelog.md` entry, no prose docs
- [ ] **If integration test modules are used:** I updated the workflow configuration — N/A
- [ ] **If this PR involves a Grafana dashboard:** I added a screenshot of the dashboard — N/A
- [ ] **If this PR involves Terraform:** `terraform fmt` passes and `tflint` reports no errors — N/A
- [ ] **If this PR involves Rockcraft:** I updated the version — N/A, no rockcraft change
- [ ] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md` — N/A, no convention change
- [ ] **If this PR changes `.copilot-collections.yaml` or `.github/instructions/`:** re-checked AGENTS.md 12-factor divergences — N/A

### DIFF
--- a/charms/garm/src/charm.py
+++ b/charms/garm/src/charm.py
@@ -25,6 +25,7 @@ from charm_state import (
     GARM_CONFIGURATOR_RELATION_NAME,
     CharmState,
     RunnerConfig,
+    configurator_related,
     credential_name,
 )
 from entity_reconciler import EntityReconciler
@@ -450,17 +451,12 @@ class GarmCharm(paas_charm.go.Charm):
 
         provider_configs = self._get_configurator_provider_configs()
         if not provider_configs:
-            # Empty provider configs are ambiguous: the garm-configurator relation
-            # may be gone, or still present but not yet publishing its
-            # secret-dependent fields (auth_url and image_id appear together once
-            # the image UUID and secrets are available). Only a removed relation
-            # leaves its scalesets orphaned and eligible for pruning; reconciling
-            # while the relation is merely mid-publish would delete live scalesets
-            # against a still-empty desired state. _reconcile_runners() is the only
-            # caller of ScalesetReconciler's orphan-delete, so run it here when —
-            # and only when — the relation is absent. GARM otherwise keeps the
-            # orphaned scalesets registered indefinitely.
-            if self.model.get_relation(GARM_CONFIGURATOR_RELATION_NAME) is None:
+            # Empty configs don't distinguish a removed relation from one still
+            # mid-publish, but only a removed relation orphans scalesets. Prune
+            # them (via _reconcile_runners) only when the relation is truly gone;
+            # reconciling mid-publish would delete live scalesets against an empty
+            # desired state.
+            if not configurator_related(self):
                 self._reconcile_runners()
             self.update_app_and_unit_status(
                 ops.WaitingStatus("Waiting for garm-configurator relation")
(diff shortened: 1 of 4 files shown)


charms/garm/src/charm.py
```diff
@@ -450,6 +451,13 @@ def restart(self, rerun_migrations: bool = False) -> None:
 
         provider_configs = self._get_configurator_provider_configs()
         if not provider_configs:
+            # Empty configs don't distinguish a removed relation from one still
+            # mid-publish, but only a removed relation orphans scalesets. Prune
+            # them (via _reconcile_runners) only when the relation is truly gone;
+            # reconciling mid-publish would delete live scalesets against an empty
+            # desired state.
+            if not configurator_related(self):
```
yanksyoon: Should this be a property in one of the charm states?

yanksyoon (commented):
Minor styling nit, thanks!

florentianayuwono (approved):

yhaliaw (approved):

weiiwang01 (approved):

charms/garm/src/charm.py
cbartz: Good call — done in 43bcc3b. It's now a `configurator_related` field on `CharmState`, populated in `CharmState.from_charm` alongside the other relation-derived state, and `restart()` reads it via `CharmState.from_charm(self).configurator_related`. Keeps the relation-presence logic in `charm_state.py` per ISD014.

---
*This reply was generated by an AI assistant.*

PR #284, opened 2026-07-14
test(garm): refactor unit tests to the scenario framework

cbartz:
#### What this PR does

Refactors the GARM charm's unit tests in `test_charm.py` to the Scenario framework, aligning
with `garm-configurator`. Each test now drives the real charm against a `State` and asserts on
the resulting `State` — secrets created, files pushed, Pebble plan, reported status — with only
GARM's HTTP surface stubbed.

The pure config-rendering and secret-generation helpers move to `test_garm_toml.py`, leaving
`test_charm.py` purely Scenario-driven.

`charm.py` statement coverage rises from 67% to 89%.

#### Why we need it

The charm-behaviour tests faked `self` with a `MagicMock` and called unbound methods
(`GarmCharm.restart(charm)`). That could only assert on the arguments the charm passed to its
collaborators — never on what it actually did — and needed scaffolding like a sentinel
exception to stop execution before `super().restart()`. Scenario removes the scaffolding and
covers the relation-parsing and config-writing paths the mock-based tests never reached.

`ops-scenario` 8.8.0 expands the `go-framework` extension itself, so no packing is needed for
the charm's container, peer relation and config options to resolve.

#### Test plan

`tox -c tox.toml -e fmt,lint,unit` — 245 unit tests pass. Coverage measured with
`coverage run --source=src -m pytest tests/unit` before and after (67% → 89% on `src/charm.py`).

#### Review focus

Rewriting the tests surfaced two pre-existing bugs. Neither is fixed here — this PR is test
work only — and both are being addressed in a follow-up that builds on these tests:

1. **The charm's Pebble layer is permanently shadowed.** `restart()` adds the `garm-command`
   layer (service `app`) before `super().restart()`. paas_charm's `_app_layer()` snapshots the
   plan into `original-services.json` on first call and re-emits those services into the `charm`
   layer, which is added later and wins. The `app` service is frozen at its first-run values, so
   proxy changes never reach the running service, and `_current_config_hash()` reads the frozen
   hash — meaning the unchanged-config short-circuit never fires again after the first config
   change. The old tests missed this by asserting on the dict passed to `add_layer` rather than
   the resulting plan.
2. **A failed GARM reconcile reports Active on the config-changed path**, because
   `paas_charm.restart()` ends with an unconditional `ActiveStatus()` after `_reconcile_runners()`
   has set `Waiting`.

Two tests changed meaning as a result, and are worth a look:

- `test_charmed_template_error_degrades_to_waiting` now exercises the steady-state
  (unchanged-config) path, the only path where the Waiting status survives (bug 2).
- The two proxy tests assert on the `garm-command` layer the charm writes, matching what the
  previous tests checked; asserting the effective plan would fail on bug 1.

Also: `restart()`'s `"Waiting for postgresql relation"` branch is unreachable — the framework's
`optional: false` gate blocks with `missing integrations: postgresql` first — so that test now
asserts the real status. And `test_restart_no_proxy_hash_matches_rendered_config` was dropped:
it recomputed the hash using the same guard as the charm, so it could not fail. The
"unchanged config does not restart the workload" transition test covers the real property.

#### New dependencies

`ops-scenario==8.8.0` added to the `unit` env in `charms/garm/tox.toml`, matching the pin
already used by `garm-configurator`. No runtime dependency change; no breaking changes.

#### Checklist

- [x] **If this PR adds/removes a charm, or changes a charm's base class, conventions, tooling, or repo structure:** I updated the relevant `AGENTS.md`

### DIFF
--- a/charms/garm/tests/unit/test_charm.py
+++ b/charms/garm/tests/unit/test_charm.py
@@ -1,15 +1,27 @@
 # Copyright 2026 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Unit tests for GarmCharm."""
+"""Unit tests for GarmCharm, driven through Scenario.
 
+Each test runs the real charm against a State and asserts on the resulting State: the Juju
+secrets it created, the files it pushed, the Pebble plan it left behind, and the status it
+reported. Only GARM's HTTP surface is stubbed (the ``garm_api`` fixture) — no GARM process
+listens in a unit test — so the charm's own logic always runs for real.
+
+The pure config-rendering helpers in charm.py are tested in test_garm_toml.py.
+"""
+
+import dataclasses
 import os
-import string
+import typing
 from unittest.mock import MagicMock, PropertyMock, patch
 
 import ops
 import pytest
 import yaml
+from ops import pebble
+from scenario import Container, Context, Model, PeerRelation, Relation, Secret, State
+from scenario.errors import UncaughtCharmError
 
 try:
     import tomllib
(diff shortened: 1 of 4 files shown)


charms/garm/tests/unit/test_garm_toml.py
```diff
@@ -0,0 +1,439 @@
...
+        "db_passphrase": "a" * 32,
+        "postgresql_config": _DEFAULT_PG_CONFIG,
+    }
+    kwargs.update(overrides)
+    toml_content, _ = render_garm_toml(**kwargs)
+    return tomllib.loads(toml_content)
+
+
+def test_render_garm_toml_postgresql_backend():
+    """The [database] section uses postgresql backend with correct fields."""
```
yhaliaw: Do you think it is worth doing the AAA pattern for the tests in this file?

yhaliaw: The tests in this file do not fit the AAA pattern.

cbartz: Yes, this needs to be fixed.

charms/garm/tests/unit/test_garm_toml.py
cbartz, line 1: these tests were just copied over to this file (and AAA docstrings added)

AGENTS.md
```diff
@@ -72,6 +72,22 @@ For **`garm-configurator`** (plain `ops`):
...
+      act: Run install.
+      assert: Both labelled secrets are created, so other units can fetch them by label.
+      """
+  ```
+  State the *why* in the assert line where it isn't obvious ("so other units can fetch them
+  by label"), not just the mechanical result. For a `parametrize`d test, write one docstring
+  covering the cases — **DON'T** write one per case.
+- **DO** fix a missing AAA docstring on any test you move or edit. It's followed unevenly
+  (`planner-operator` and `garm` yes; `garm-configurator` and `webhook-gateway-operator` not
+  yet), so imitating the nearest neighbour is not a reliable guide.
```
cbartz, line 90: I see AGENTS missing the AAA pattern regularly, thus I've added this

charms/garm/tests/unit/test_charm.py
cbartz, line 1: the PR description has a test migration plan overview - the intent of the previous tests has been preserved

charms/garm/tests/unit/test_charm.py
```diff
@@ -1,1418 +1,938 @@
...
+    assert: No credential is built — GARM rejects a partial one.
     """
-    charm = _github_charm([{"github_app_id": "1"}])
+    state = _state(configurator_units_data={0: {**_PROVIDER_UNIT_DATA, **github_data}})
 
-    credentials = GarmCharm._build_desired_credentials(charm)
+    ctx.run(ctx.on.update_status(), state)
 
-    assert credentials == []
+    garm_api.github.return_value.reconcile.assert_called_once_with([])
```
florentianayuwono, line 811: where is the assertion for no credential is built?

florentianayuwono: oh wait the `[]` signify no credential is it

